PR #26, opened 2026-04-04
fix: harden AgentMember config/schema deserialization against legacy records

randlee:
## Summary

- Adds persisted-data repair guide covering config/schema recovery policy
- Defines recovery rules: compatibility-only drift gets deterministic defaults; identity/routing fields remain required; non-recoverable failures preserve file/entity/parser context
- Links repair guide from product config-recovery requirement and atm-core architecture docs

This is the docs-first increment. The serde fix (`#[serde(default)]` on `agent_type` + regression test) follows in a subsequent commit on the same branch.

**Fixes**: `missing field agentType` regression reported by jen@src-gen after `agentType` was added as a required field to `AgentMember` without a serde default.

## Test plan

- [ ] Docs-only increment — no code changes, no tests required
- [ ] Serde fix + regression test in next commit on this branch

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -47,7 +47,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
- "windows-sys",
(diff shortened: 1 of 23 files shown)


randlee:
Superseded by #27 (broader hardening scope). Same branch — closing duplicate.

PR #190, opened 2026-05-07
fix(qa): require full workspace pattern sweep on every repeatable finding

randlee:
## Summary
- QA step e updated: any finding that is a repeatable pattern must include a full workspace search for all occurrences before the finding is reported
- Fixes the root cause of multi-round QA churn where a pattern is fixed in one place and re-found in 2-3 others in later rounds

## Test plan
- [ ] Verify qa-template renders correctly with existing var files


randlee:
Wrong base: was develop→main. Reopening as feature/qa-template-pattern-sweep→develop.

PR #285, opened 2026-05-15
feat: complete phase X guardrails and closeout

randlee:
## Summary
- add Phase X closeout guardrail scripts and wire them into local lint/CI entrypoints
- move retained-runtime bootstrap ownership out of the CLI production dependency line and into bootstrap/test-support helpers
- finalize X.5 sprint/project-plan documentation and stabilize retained-runtime tests for spawned-thread and multi-fixture coverage

## Validation
- cargo fmt --all --check
- python3 .just/run_lint.py all
- cargo build --workspace
- cargo test --workspace
- git diff --check


randlee:
## QA-1 Verdict — FAIL 1B+8I+8m

**Branch**: feature/pX-s5-guardrails-and-closeout @ 78d1e2c
**Sprint**: X.5 — Guardrails, Dependency Ownership, And Closeout Verification
**Round**: QA-1

---

### Sprint Gate ✅

- `docs/phase-X/sprint-X5.md` exists, `status: complete`, `branch:` set, `worktree:` set
- `docs/project-plan.md` includes X.5 sprint entry
- `scripts/check-legacy-mailbox-paths.py` and `scripts/check-capability-degradation.py` present, runnable, wired into Justfile and `.just/run_lint.py` ✅
- cargo-shear wired via `lint_cargo_shear.py` in `CARGO_LINT_ORDER` ✅
- TASK-1515 baseline: all 4 artifacts confirmed present and consistent ✅
- `atm/Cargo.toml` no `atm-rusqlite` in `[dependencies]` ✅
- `atm-daemon-bootstrap` installs production factory correctly ✅
- Boundary files updated (`atm-protocol.toml`, `sqlite-boundary-assembly.toml`) ✅
- `check-function-length.py` trait-method exclusion fix confirmed working ✅
- `cargo fmt` PASS, `cargo clippy` PASS, 578+ tests PASS ✅

---

### BLOCKING (1)

**ARCH-001** — Module-level `#[cfg(all(test, unix))]` gate (Phase S boundary policy violation)
`crates/atm-daemon/src/composition.rs:763`
The entire test module is gated with `#[cfg(all(test, unix))]`. The Phase S boundary policy in `docs/atm-daemon/boundaries.md` explicitly states: module-level platform test gates are not allowed in daemon-owned test modules; use `#[cfg(unix)]` on individual test functions that require Unix semantics. Platform-neutral lifecycle and shutdown tests are silently excluded on Windows.
Fix: replace `#[cfg(all(test, unix))]` on the mod declaration with `#[cfg(test)]`. Add `#[cfg(unix)]` on individual test functions that actually require Unix semantics.

---

### IMPORTANT (8)

**RBP-F001 / RSH-X5-001** — `install_default_runtime_factory` silently discards duplicate install (carry-forward from X.1 QA-2, unfixed)
`crates/atm-core/src/service_runtime_store.rs:14`
`let _ = DEFAULT_RUNTIME_FACTORY.set(factory)` discards the `OnceLock::set` `Err` with no log entry or error propagation. Second caller gets no signal that their factory was dropped.
Fix: emit `tracing::warn!` when `set` returns `Err`, or return `Result` so callers can detect the conflict.

**RBP-F002** — `AckRequest::reply_body: String` raw/unvalidated on public struct
`crates/atm-core/src/ack/mod.rs:28`
Validation via `input::validate_message_text` is deferred deep into `ack_mail_with_runtime_sqlite`. Both public entry points (`ack_mail`, `ack_mail_with_runtime`) accept unvalidated body. Contrast with `ReadQuery::new` and `SendRequest::new` which validate at construction time.
Fix: introduce `AckRequest::new` that validates `reply_body` at construction, or move the validation into a validated newtype.

**RSH-X5-002** — `thread::sleep` in `wait_for_eligible_message` blocks IPC handler thread (carry-forward from X.1 QA-2, unfixed)
`crates/atm-core/src/read/wait.rs:30`
100ms `thread::sleep` runs inline on the path reachable from the daemon's IPC handler. A read-with-timeout pins the handler OS thread for the full poll interval per iteration.
Fix: add doc comment confirming whether this is safe from IPC context; wrap in dedicated thread or file-watch notification if reached from daemon handler.

**FTQ-001** — `SqliteRuntimeGuard::install` mutates env var without `env_lock()` — racy under parallel tests
`crates/atm-runtime-test-support/src/lib.rs:31`
`set_env_var` is called without acquiring the shared `env_lock()` Mutex. Two concurrent `Fixture::new()` calls in parallel test threads can race on `ATM_TEST_SQLITE_RUNTIME_PATH` read-modify-write, causing wrong-path restoration on drop.
Fix: acquire `env_lock()` before reading previous value and writing new value in `install`, and hold it through `Drop`.

**FTQ-002** — `SQLITE_RUNTIME_CACHE` path collision risk under FTQ-001 race
`crates/atm-runtime-test-support/src/lib.rs:63`
Cache keyed by `PathBuf` read from env var. If FTQ-001 allows one test to see another test's path, it may receive a cached runtime built for a different SQLite database, silently corrupting assertions.
Fix: address FTQ-001 first; additionally consider per-test fresh assembly rather than path-keyed caching.

**FTQ-003** — Spin-without-sleep in `wait_for_missing_config_notice` / `wait_for_workflow_state_for_message`
`crates/atm-core/tests/mailbox_locking.rs:977`
`thread::yield_now()` loop makes N filesystem reads per second competing with writer threads. Under CI load this starves writers and exhausts the 30s deadline, producing misleading "not observed before timeout" panic.
Fix: replace `yield_now()` with a bounded sleep (1-5ms) or channel-based completion signal.

**FTQ-004** — Dual `Once`/factory pairs race for the same global singleton
`crates/atm-daemon/src/tests.rs:41` and `crates/atm-runtime-test-support/src/lib.rs:12`
Both call `atm_core::install_default_runtime_factory`. If both are linked in the same test binary, whichever `Once` fires first wins; the other is silently dropped. Tests depending on the losing factory get daemon-unavailable errors.
Fix: consolidate to one factory per binary — have `atm-daemon` use `atm-runtime-test-support` instead of re-implementing its own `Once`/factory pair.

**FTQ-005** — Non-`#[serial]` tests call `install_retained_runtime_factory()` in `atm-daemon/tests.rs`
`crates/atm-daemon/src/tests.rs:358`
`heartbeat_updates_status_cache_and_doctor_projection` and several other tests call `install_retained_runtime_factory()` and set `ATM_TEST_SQLITE_DB_PATH` via `EnvGuard` without `#[serial]`. They can race with `#[serial]` tests that mutate the same env var.
Fix: mark all tests calling `install_retained_runtime_factory()` with `#[serial]`, or migrate to `SqliteRuntimeGuard` after fixing FTQ-001.

---

### MINOR (8)

**RBP-F003** — `SQLITE_RUNTIME_CACHE` OnceLock<Mutex<...>> lacks rationale comment
`crates/atm-runtime-test-support/src/lib.rs:13`
Interior mutability pattern requires inline rationale per RBP-006. Add 2-3 line comment explaining why shared mutation across test threads is needed.

**RSH-X5-003** — `resolve_daemon_bin` reads `ATM_DAEMON_BIN` lazily at call time
`crates/atm-daemon-bootstrap/src/lib.rs:31`
No startup-time validation that the path exists/is executable. Misconfigured path fails at first use, not at startup.

**RSH-X5-004** — `WORKFLOW_LOCK_TIMEOUT` not operator-configurable
`crates/atm-core/src/service_runtime.rs:13`
Fixed 5s constant with no config knob; add a comment documenting the rationale and target deployment environment.

**FTQ-006** — Fixed 1ms sleep in `connect_daemon_local_ipc_until_ready` retry loop
`crates/atm-daemon/src/test_support.rs:112`
Flat backoff spins up to 10,000 times over 10s; use exponential backoff capped at ~50ms.

**FTQ-007** — 10s wall-clock deadline for `wait_for_message_contains` observability assert
`crates/atm-daemon/src/tests.rs:248`
Hard wall-clock deadline can flake under CI parallelism; increase or switch to condvar-based wait.

**FTQ-008** — Socket path read from `HOME` env var outside serialized section
`crates/atm-daemon/src/tests.rs:203`
`daemon_socket_path()` resolved from `HOME` before the critical section; non-serialized tests may mutate `HOME` concurrently. Use a `TempDir`-scoped socket path.

**ATM-QA-X5-001** — `ALLOWED_TEST_MATCHES` variable name misleading in `check-capability-degradation.py`
`scripts/check-capability-degradation.py:23-28`
Variable name implies test-only carve-outs but contains a production file (`peer_transport.rs`). Rename to `KNOWN_SAFE_CARVEOUTS` and add inline comment explaining why the production occurrence is safe.

**ATM-QA-X5-002** — Overly broad `legacy:01` regex fragment in `check-legacy-mailbox-paths.py`
`scripts/check-legacy-mailbox-paths.py:46-50`
Short literal `legacy:01` in `ALLOWED_LEGACY_LITERALS` alternation can silently suppress future regressions. Replace with a more precise pattern or add a doc comment for each alternation branch.

---

### Fix Priority

1. ARCH-001 — remove module-level `#[cfg(all(test, unix))]` from `composition.rs:763`
2. FTQ-001 + FTQ-004 — env locking in `SqliteRuntimeGuard` + consolidate to one factory per binary
3. FTQ-002 — address path collision after FTQ-001
4. FTQ-003 + FTQ-005 — yield_now() spin + missing `#[serial]`
5. RBP-F001 / RSH-X5-001 — silent factory discard (carry-forward — needs final resolution)
6. RSH-X5-002 — thread::sleep IPC handler path (carry-forward)
7. RBP-F002 — `AckRequest::reply_body` validation at construction
8. Minors as available

randlee:
Superseded by Phase Xb restart. Salvage cherry-picks applied to integrate/phase-Xb. Closing per team-lead authorization.

PR #571, opened 2026-07-17
fix(cross-host): route ACK replies back to origin host (AG-FIND-005) + AG-VAL-022 Win/macOS smoke

ehrvs:
## Summary

Fixes **AG-FIND-005** for the Windows↔macOS pair: cross-host `--requires-ack`
messages now complete a real ack round-trip. Discovered and validated by running the
**AG-VAL-022** heterogeneous smoke matrix live on two real hosts (macOS `192.168.1.178`
↔ Windows `192.168.1.146`, secure-required pinned-fingerprint TLS).

## Critical findings

**AG-FIND-005 (blocking) — cross-host ack replies never returned to the origin host.**
Two independent layers each dropped the origin:
1. **Ack path never routed remotely.** The inbound peer origin was discarded at
   dispatch (`peer_transport/server.rs`), the durable message model had no origin-host
   field, `ack::ReplyTarget` could not represent a host, and the ack reply was
   hard-wired to a *local* inbox path — so a Windows ack to a Mac-origin message landed
   in a Windows-local same-named mailbox (or errored) instead of returning to the Mac.
2. **SQLite writer dropped the field.** Even after (1), the `atm-storage-rusqlite`
   writer serializes via a hand-maintained `StorageEnvelope` projection that enumerates
   fields explicitly and had not been updated for `origin_host`, so it was silently
   dropped on every write and reloaded as `None` at ack time. **This was invisible to
   unit tests and only surfaced under live cross-host validation.**

**AG-FIND-006 (non-blocking, opened, not fixed here) — downed-peer send error quality.**
A cross-host send to a stopped peer surfaces as a local
`failed to read daemon response frame (os error 35)` rather than a clear
remote-unreachable / retry-budget-exhausted classification.

## Fix

- Persist inbound peer origin as `MessageEnvelope.origin_host` (serde `originHost`,
  `default` + `skip_serializing_if` — legacy JSON → `None`, **no migration**; storage is
  a JSON blob in the SQLite envelope column).
- Thread peer origin through `RequestDispatcher::dispatch(peer_origin)`; the peer TCP
  listener passes the connecting IP, the same-host IPC listener passes `None`.
- Extend `ack::ReplyTarget` with `Option<RemoteTargetHost>`, resolved from the source
  message's `origin_host`.
- Split `persist_sent_ack_reply`: origin `None` → unchanged local path; `Some` → build a
  `SendRequest` (with `remote_host`, `acknowledges_message_id`) and route it via the new
  sealed `boundary::AckReplyDeliveryPort`, backed in the daemon by
  `cross_host_delivery.deliver_remote` (same outcome mapping as compose-send; no silent
  local fallback on remote failure).
- `persist_source_ack_state` runs in both branches, so the source message is marked
  acknowledged regardless of reply routing.
- Fix the `StorageEnvelope` write projection to carry `origin_host`.

Local / same-host ack behavior (`origin_host = None`) is unchanged; **AG-VAL-017** stays
green.

## Validation (live, real hosts)

| Row | Result |
|---|---|
| AG-VAL-022A unauthorized rejection | PASS (rejected before mutation, exit 3) |
| AG-VAL-022B authorized durable send (both directions) | PASS |
| AG-VAL-022C receiver read | PASS (`mutation_applied: true`) |
| **AG-VAL-022D cross-host ack round-trip** | **PASS (post-fix)** — receiver stores `originHost=192.168.1.178`; ack `reply_target=arch-mac@…​.192.168.1.178`; origin Mac inbox receives the reply with `acknowledgesMessageId` set |
| AG-VAL-022E degraded-notification | covered-by-design |
| AG-VAL-022F retry-visible recovery | PASS (+ AG-FIND-006 note) |

Before the fix, 022D delivered the reply to the receiver-local mailbox (confirmed via
code trace + live run). After the fix, the reply round-trips to the origin host.

## Tests

- `schema::inbox_message` origin_host serde round-trip incl. legacy-JSON-without-originHost → None.
- `ack` remote path routes a `SendRequest` with `remote_host` set via a mock
  `AckReplyDeliveryPort` and marks the source acknowledged with no local write.
- `atm-storage-rusqlite` origin_host SQLite write/read round-trip + legacy-row → None.
- Full workspace: `cargo build/test --workspace`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo fmt --all --check` — all clean on macOS and built clean on Windows (`x86_64-pc-windows-msvc`).

## Notes for review

- Storage is JSON-blob-backed, so **no SQLite schema migration** was required.
- `RequestDispatcher::dispatch` signature changed (compile-forced sweep across all impls
  incl. tests) to guarantee no call site silently drops the origin.
- `cargo fmt` reformatted some pre-existing long lines in `peer_transport/delivery.rs`.
- Docs updated on this branch: `cross-host-smoke-checklist.md` (022A-F → PASS) and
  `cross-host-findings-ledger.md` (AG-FIND-005 validated, AG-FIND-006 added).


ehrvs:
### Critical review complete — all findings resolved and re-validated live

Ran rust-code-reviewer, rust-best-practices, and boundary-guard over the diff, plus `cargo test -p atm-architecture` (8/8) and full-workspace build/test/clippy `-D warnings`/fmt (clean on macOS; built clean on Windows `x86_64-pc-windows-msvc`).

**Findings addressed (commit 74360870):**
- **RCR-001 (important, correctness)** — the remote ack path fabricated a `reply_message_id` that was never transmitted, so the id reported to the caller didn't match the delivered message. `AckReplyDeliveryPort::deliver_reply` now returns the real remote-assigned id (extracted from the `Delivered` `ResponseEnvelope`; non-`Sent` shapes are a hard error, never a fabricated id).
- **RBP-F001 (important, type safety)** — `SendRequest.origin_host` retyped `Option<String>` → `Option<RemoteTargetHost>`; the daemon parses `peer_origin` via `RemoteTargetHost::parse` (errors propagate).
- **RBP-F002 (minor)** — storage stays `Option<String>` at the atm-storage boundary (below atm-core; no forbidden edge / no migration); validation centralized at `resolve_reply_target` with a fail-closed test.
- **boundary-guard: PASS** — layering intact, `AckReplyDeliveryPort` sealed + daemon-injected, `RemoteTargetHost` core-owned, no forbidden edges.

**Live re-validation (macOS 192.168.1.178 ↔ Windows 192.168.1.146, secure-required pinned-TLS):**
- Original send `message_id = 01KXS7VE8GB24KC4AQDSSXQ047`
- Windows ack reported `reply_message_id = 01KXS7WES57CA3A2JE70A8WT8R`, `reply_target = arch-mac@xhost-smoke.192.168.1.178`
- Mac origin inbox received `message_id = 01KXS7WES57CA3A2JE70A8WT8R` (**exact match**), `acknowledgesMessageId = 01KXS7VE8GB24KC4AQDSSXQ047`

AG-VAL-022A–F all pass (022E covered-by-design; 022F pass + AG-FIND-006 error-quality note filed). Ready for review — @randlee.

rand-lee:
Reviewed against the current baseline on `feature/pAG-s15-othermac-smoke` at `f806cbeb`.

My read is that the main useful part of this PR is already in the baseline, while 2 real bugs remain in the branch.

1. ACK/host support is already landed in the base line
- We already took the cross-host ACK routing fix in `e2e81ae7` (`Fix ack reply routing for remote hosts`).
- That baseline change already touched the same core surfaces this PR is trying to improve: `crates/atm-core/src/ack/mod.rs`, `crates/atm-core/src/send/context.rs`, `crates/atm-daemon/src/runtime_health/dispatch_delivery.rs`, `crates/atm-daemon/src/peer_transport/server.rs`, and the peer-transport delivery path.
- Current AG.14 / AG.15 also already have passing local coverage around remote-target send/ack paths (`crates/atm-daemon/src/tests/runtime_root/loopback.rs`, `crates/atm-daemon/src/tests/runtime_root/self_ip.rs`) plus the AG.14 merge-forward at `f806cbeb`.
- So I do not see a net-new capability here that the current base still lacks.

2. Dot-notation ambiguity is still a real bug in this branch
- This branch serializes/deserializes reply targets as `<agent>@<team>.<host>` in `crates/atm-core/src/ack/mod.rs` (the `ReplyTarget` serde logic splits `team_and_host` on `.`).
- But the current type validators still allow `.` inside agent/team names at `crates/atm-storage/src/validation.rs:28-31`:
  `ch.is_ascii_alphanumeric() || matches!(ch, '-' | '_' | '.')`
- That means the branch is depending on a no-dots invariant that is not actually enforced by the current code path. Until that invariant is enforced first, this encoding is ambiguous.
- We do have the exact remote-target contract documented in ADR/plan docs (`docs/adr/ADR-031-remote-target-contract-and-cross-host-dispatch.md`, `docs/requirements.md`, `docs/architecture.md`), but the branch still needs the validator side to match that contract.

3. Failed-ack gap is still a real bug in this branch
- In this branch, `persist_ack_reply()` first calls `persist_source_ack_state()` and only after that attempts remote reply delivery.
- Concretely:
  - local ack state is persisted first in `crates/atm-core/src/ack/mod.rs` (`persist_ack_reply` -> `persist_source_ack_state`)
  - remote reply is attempted later in `persist_remote_sent_ack_reply`
- Then `DaemonRequestDispatcher::deliver_reply()` maps deferred / outcome-unknown remote reply delivery to hard errors in `crates/atm-daemon/src/runtime_health/dispatch_delivery.rs`.
- Net effect: `atm ack` can return an error after the local source message has already been marked acknowledged and removed from the retryable pending-ack path.
- That is worse than baseline behavior because the operator sees a failed ack command but the local ack has already been consumed.

Bottom line:
- useful ACK/host support: already in base (`e2e81ae7` + AG.14/15 line)
- remaining branch bugs: dotted-target ambiguity + failed-ack-after-local-consume gap
- conclusion: I would not merge this PR as-is

ehrvs:
Closing as **superseded** — thanks @rand-lee, your review is correct.

Reconciling against the current baseline (`feature/pAG-s15-othermac-smoke`, now `328993e1`) confirms:

- The cross-host ACK routing this PR implements is **already landed** in `e2e81ae7` + the AG.15 follow-ups, via a cleaner metadata-based mechanism (`remoteHost` in the message `extra` map, threaded through the normal send path). This PR's typed-field + dedicated-`AckReplyDeliveryPort` design is redundant.
- The SQLite `origin_host` projection fix here only repairs a regression this PR's own typed-field approach introduced; baseline carries the host in the flattened `extra` map, which round-trips naturally — not net-new.
- The dot-notation `<agent>@<team>.<host>` ambiguity exists identically in baseline (shared design), so the encoding here is not a fix.
- This branch additionally has a **failed-ack-after-local-consume regression** (deferred/outcome-unknown mapped to a hard error after `persist_source_ack_state`) that baseline does **not** have — baseline maps those to success.

Net: no capability here that baseline lacks, plus a regression. The one implementation-agnostic net-new item — **AG-FIND-006** (unreachable-peer send surfaces as a local `failed to read daemon response frame (os error 35)` rather than a clear remote-unreachable/retry-budget classification) — has been captured separately on `feature/pAG-ack-reconcile` (docs-only, off current baseline).

Separately, for your consideration (not in this PR): hardening `crates/atm-storage/src/validation.rs` to forbid `.` in agent/team names would remove the shared dotted-target ambiguity in the reply-target encoding.

Live cross-host smoke (macOS↔Windows, secure-required pinned-TLS) was still valuable as validation of the end-to-end path and surfaced AG-FIND-006; I'll re-run it against the baseline mechanism.

PR #809, opened 2026-08-09
feat(skills): add hermes-gateway status/reset skill

randlee:
## Summary

Adds a repo-local `hermes-gateway` skill for checking status of, or resetting,
a single named Hermes-graft agent's gateway.

- Vendors `hermes_gateway` (originally authored by the `skillrx` Hermes-graft
  agent for this skill) at `.claude/skills/hermes-gateway/scripts/hermes_gateway`
  rather than assuming it's a bare command on `PATH`.
- Mandatory `agent_name` parameter — no fleet-wide or unscoped operation.
- Default invocation (`/hermes-gateway <agent>`) is status-only; `--reset`
  triggers the full endpoint-reset procedure. Status is always checked first,
  even on a `--reset` invocation, before any reset action is taken.
- Trigger phrases: "restart hermes gateway", "hermes gateway", "hermes
  gateway status", "reset hermes gateway".

## Test plan

- [ ] `python3 -m py_compile` on the vendored script (done locally, passes)
- [ ] arch-ctm / skillrx@hermes review
- [ ] arch-ctm mirrors this as a `.codex` skill on the same worktree

### DIFF
--- a/.claude/skills/hermes-gateway/SKILL.md
+++ b/.claude/skills/hermes-gateway/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: hermes-gateway
-description: Check status of, or reset, a Hermes graft agent's gateway. Use when a Hermes-graft agent's published endpoint record is stale (wrong schema_version, wrong owner_generation, or workspace-rooted instead of roster-home-rooted) and sends are accepted but the receiver hook fails to decode or route them, or when simply checking whether a profile's gateway is up. Triggers: "restart hermes gateway", "hermes gateway", "hermes gateway status", "reset hermes gateway".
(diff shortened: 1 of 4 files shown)


randlee (commented):
## skillrx review

### ✓ SKILL.md — approved with minor notes

The reset procedure (preconditions, rules, poll logic) correctly reflects the graft bridge lifecycle: endpoint-root symlink repair → gateway restart via `launchctl kickstart -k` → authenticated live-proof send. Rules 1–3 (no hand-editing, no synthetic events, no self-restart) are exactly right.

**One nit on polling (step 5):** `launchctl kickstart -k` with KeepAlive in the plist brings the gateway back in ~3–5s. Six 5-second polls is conservative and fine, but you could tighten to ten 2-second polls and catch the new PID sooner.

### ✗ Vendored script is stale — needs update from SSOT

`.claude/skills/hermes-gateway/scripts/hermes_gateway` is an **older snapshot** of the script. The canonical version at `hendrix/scripts/hermes_gateway` (symlinked to `~/bin/hermes_gateway`) has since gained:

- `$HERMES_HOME` env var support with auto-detection of profile-vs-base path
- `--json` output mode
- `gateway_info()` data consolidation (single source of truth for all profile data)

The SSOT is `hendrix/scripts/hermes_gateway` — copy from there, not from `~/bin/` or any profile-local copy.

### Status

Approve after updating the vendored script from the hendrix SSOT. SKILL.md content is good.

PR #896, opened 2026-08-16
fix(hermes-atm): RC verification blockers (installer toolset, non-mutating read, CLI read selection)

randlee:
## Summary

Fixes 3 release-blocking findings from live hermes-atm RC verification against `develop`, independently corroborated by GitHub issues filed from real release-proof testing (alpha-prime/omega-prime@hermes).

- **#894 / HERMES-RC-001**: `hermes-atm`'s installer only wrote `plugins.enabled` in Hermes' config, not the required `platform_toolsets` entry -- left the plugin manifest "enabled" while native tools were functionally absent (omega-prime profile). Commit `1e9aca55`.
- **#893 / HERMES-RC-002**: native `atm_read` (via `read_tool`) used the mutating read path (`read_mail_with_runtime_impl(MutatingRead)`) instead of the non-mutating peek path, violating atm_read's documented "without mutating it" contract. Commit `cd59d6b7`.
- **#895 / HERMES-RC-003**: CLI `atm read`'s `with_caller_chat_id()` set a `participant_filter` requiring `destination_chat_id == chat_id`, but agent-addressed mail has NULL `destination_chat_id` -- silently dropped all unread mail from CLI `atm read` (both `--unread` and default selection). This is core CLI behavior, not hermes-atm-specific. Commit `a34d4840`.

## Test plan
- [x] `cargo fmt --check`, `cargo clippy --workspace -- -D warnings`, `just lint`, full `cargo test`, `just test` (506 tests)
- [x] Clean `hermes-graft` bridge wheel install
- [x] Fresh rebuilt-wheel reinstall/reset live proof: native tools registered; native read `mutation_applied=false` and message stayed unread; native send delivered (with one transient reconnect retry)
- [ ] quality-mgr QA pass

PR #897, opened 2026-08-16
fix(release): align hermes-atm version with workspace

randlee:
## Summary

Found during post-merge live verification of PR #896: `hermes-atm`'s own package version was out of sync with the rest of the workspace (which was contributing to the deployment-staleness confusion seen during Alpha/Omega retesting).

- `hermes-atm` bumped to 1.4.2, matching the workspace and `atm-graft`.
- Added CI coverage (`cargo test -p agent-team-mail --test version_alignment`) that locks all shipped Python/native package manifests to the workspace version, so this can't drift silently again.

Commit: `1afe4b171`

## Test plan
- [x] `cargo test -p agent-team-mail --test version_alignment` (2/2)
- [x] Isolated clean-wheel bridge test (17/17)
- [ ] quality-mgr QA pass


randlee:
# QA-HERMES-VERSION-ALIGN-1 — QA-1 for PR #897 (chore/hermes-atm-version-alignment)

**Branch:** `chore/hermes-atm-version-alignment` @ `1afe4b17` (target: `develop`)
**Verdict:** **FAIL**
**Reviewers returned:** 5/5

```json
{
  "sprint": "hermes-atm-version-alignment",
  "task": "QA-HERMES-VERSION-ALIGN-1",
  "branch": "chore/hermes-atm-version-alignment",
  "commit": "1afe4b17",
  "pr": 897,
  "verdict": "FAIL",
  "findings": { "blocking": 1, "important": 2, "minor": 3 },
  "blocking_ids": ["RBQA-F001"],
  "next_action": "Fold the new version check into .just/check_version_sync.py (the existing canonical owner) instead of duplicating it in Rust with different, conflicting semantics",
  "owner": "arch-ctm",
  "merge_readiness": "not ready",
  "merge_reason": "New test will conflict with the existing CI-wired Python version-sync check the next time the workspace version carries a prerelease qualifier -- a real, historically-used pattern in this repo, not hypothetical"
}
```

## Blocking — found by ruthless-boundary-qa, independently confirmed by me

**RBQA-F001 (critical):** This repo already has a canonical, CI-wired owner for "does a shipped Python/native package manifest's version match the workspace version" — `.just/check_version_sync.py`, registered as the `"version"` lint task in `.just/run_lint.py:100` and wired into `just lint` via the Justfile's `_lint-version` recipe. Its `validate_python_graft_version` deliberately normalizes the workspace version to its numeric base (`normalized_version_base`, stripping any `-prerelease`/`+build` suffix) before comparing against `atm-graft-python/Cargo.toml` — because during a prerelease workspace version, that manifest is *required* to stay at the plain numeric base for PEP 440 compatibility. This is proven intentional, not an oversight: there's a dedicated test (`test_python_graft_version_strips_prerelease_workspace_qualifier`) and real historical precedent — `.triage/phase-AI/findings/AI25-QA-003-RESIDUAL.ttl` documents a real production release at `1.3.2-beta.27`.

The new Rust test (`released_python_artifact_versions_match_workspace_version`) instead does a **raw string equality** comparison with no prerelease normalization, requiring the *same file* to contain the *full* prerelease string. **The two checks cannot both pass simultaneously the next time a prerelease workspace version is cut.** I independently verified this by reading `check_version_sync.py` directly: `normalized_version_base`'s regex captures only the `\d+\.\d+\.\d+` group, discarding everything after `-`/`+`, and confirmed the historical `1.3.2-beta.27` precedent is real, not a hypothetical.

I also confirmed the mutation-test result the other reviewers relied on (rust-qa-agent's live `cargo test` run, which genuinely failed on a version mismatch) is correct for the **current** plain `1.4.2` workspace version — but that result doesn't cover the prerelease-conflict scenario, since no reviewer tested with a prerelease workspace version. arch-qa and req-qa both flagged (correctly, as a non-blocking caveat) that they lacked shell access to run the mutation test themselves; ruthless-boundary-qa's workspace-wide duplicate-implementation sweep is what caught the actual design conflict that a mutation test on the current version wouldn't surface.

**Recommendation** (ruthless-boundary-qa's, which I agree with): delete the new test's manifest-version checks and instead extend `check_version_sync.py` to cover the two currently-uncovered `pyproject.toml` files (`atm-query-python`, `hermes-atm`), reusing its existing prerelease-stripping rule, so there is exactly one owner and one correctness rule for this decision. If Rust-side coverage is still wanted, it should assert on `check_version_sync.py`'s behavior rather than reimplementing the comparison independently.

## Important

- **RBQA-F002**: the pre-existing `atm_core_dep_version_matches_workspace_version` function (extended, not newly authored, by this PR) also duplicates decision logic `check_version_sync.py`'s `validate_crate_versions`/`validate_lockfile` already cover more comprehensively (for every workspace member, not just `agent-team-mail`), using hand-rolled parsing instead of the existing `tomllib`-based approach.
- **RBP-F001** (rust-best-practices-agent): the two test functions in the same file independently reimplement identical manifest-version-line parsing logic instead of sharing the file's own `manifest_version()` helper.

## Minor

- **RBP-F002/F003**: panic messages don't explain the literal-version-vs-`version.workspace = true` constraint, and recovery guidance is inconsistent across the file's own assertions (one tells you to run `cargo generate-lockfile`, others just state the mismatch).
- **QA-001** (rust-qa-agent): `manifest_version()` uses naive line-prefix string matching rather than a real TOML parser — currently safe for all 6 manifests checked, but convention-dependent rather than structurally guaranteed.
- (Noted by rust-best-practices-agent, not filed as a finding) the pre-existing dep-version lookup in `atm_core_dep_version_matches_workspace_version` uses `.find()` against a file where `agent-team-mail-core` appears twice (deps + dev-deps), so it only ever checks the first occurrence.

## What verified cleanly

- **hermes-atm's version bump itself**: genuinely `1.4.2`, matching the workspace version, confirmed independently by all 5 reviewers plus my own check.
- **Manifest-coverage completeness**: confirmed independently by all 5 reviewers — the new test's 5-manifest/3-package enumeration is exhaustive against the current workspace (no shipped Python/native manifest is missed). This part of the task's verification checklist is fully satisfied; the coverage-*breadth* claim was accurate. The problem is the check's *architectural placement and comparison semantics*, not its coverage.
- **Live mutation test**: rust-qa-agent actually executed it (bumped `hermes-atm/pyproject.toml` to `1.4.1` in a scratch edit, ran `cargo test -p agent-team-mail --test version_alignment`, confirmed a clear failure message, reverted cleanly). Valid for the current plain-version case; doesn't cover the prerelease-conflict scenario above.
- Full workspace `cargo fmt --check`, `clippy -D warnings`, `cargo test` all pass.

## Recommendation

Do not merge as-is. Fold the new check into `.just/check_version_sync.py` (extending its manifest coverage to `atm-query-python`/`hermes-atm` and reusing its existing prerelease-normalization rule) rather than duplicating the decision in Rust with conflicting semantics. This is a small, well-scoped fix given the existing infrastructure already does the hard part correctly.

randlee:
# QA-HERMES-VERSION-ALIGN-2 — Recheck for PR #897

**Branch:** `chore/hermes-atm-version-alignment` @ `449c641a` (an unrelated, unauthorized commit `eb422c1d0` also landed on this branch mid-review — team-lead confirmed it out of scope; verified it touches none of the review targets)
**Verdict:** **FAIL**
**Reviewers returned:** 4/4

```json
{
  "sprint": "hermes-atm-version-alignment",
  "task": "QA-HERMES-VERSION-ALIGN-2",
  "branch": "chore/hermes-atm-version-alignment",
  "commit": "449c641a",
  "pr": 897,
  "verdict": "FAIL",
  "findings": { "blocking": 0, "important": 2, "minor": 2 },
  "next_action": "Delete atm_core_dep_version_matches_workspace_version and rely on check_version_sync.py's validate_crate_versions/validate_lockfile as sole owner, same remediation pattern already applied successfully to RBQA-F001",
  "owner": "arch-ctm",
  "merge_readiness": "not ready",
  "merge_reason": "RBQA-F001 (critical) is genuinely resolved; RBQA-F002 (important) is not -- the fix changed comparison semantics without collapsing duplicate ownership, introducing a new false-negative gap"
}
```

## RBQA-F001-VERSION-CONFLICT — RESOLVED (unanimous, 4/4)

Fully and independently confirmed by all 4 reviewers: the duplicate/conflicting Rust manifest-alignment test is genuinely deleted (zero remaining references anywhere in the repo), and `check_version_sync.py`'s single-manifest check was generalized into `validate_python_release_versions`, iterating all 5 released-Python-artifact manifests through the one shared `normalized_version_base()` prerelease-stripping rule. Backed by generalized unit tests plus a real-repo integration test. This is the correct, complete fix for the original critical finding.

## RBQA-F002 — NOT RESOLVED (found by ruthless-boundary-qa, independently corroborated by rust-qa-agent from a different angle)

**ruthless-boundary-qa's finding, which I independently verified by reading the code directly:** the fix changed the retained `atm_core_dep_version_matches_workspace_version` Rust test's comparison from raw string equality to `release_version_base()`-based stripped-base equality — but did **not** collapse it into `check_version_sync.py`'s existing `validate_crate_versions`/`validate_lockfile`, which still independently own the same decision (does the `crates/atm/Cargo.toml` `agent-team-mail-core` path-dep pin, and Cargo.lock, match the target crate's version) for the entire workspace. Worse, the change introduced a **new** semantic gap: `crates/atm-core/Cargo.toml` uses `version.workspace = true`, so Python's `expected_package_version()` resolves atm-core's expected version to the **raw, unstripped** workspace version, and `validate_crate_versions`/`validate_lockfile` do **exact raw-string comparison with zero normalization**. I confirmed this directly in the code. So during a prerelease workspace version, if the `crates/atm/Cargo.toml` pin isn't bumped to match exactly, Python's canonical `just lint` gate correctly fails — but the Rust test's new stripped-base comparison would incorrectly pass, giving false confidence in local `cargo test` runs.

**rust-qa-agent independently corroborated the same underlying gap** by actually executing the prerelease mutation scenario (`1.4.2-beta.1`, scratch edit, reverted): `validate_python_release_versions` (the function 449c641a targeted) passed correctly in isolation, but the *full* `check_version_sync.py` run failed in the untouched `validate_crate_versions` — exactly the exact-comparison, no-normalization behavior ruthless-boundary-qa flagged — while `cargo test` itself couldn't even reach the target tests because Cargo's own dependency resolver rejects `atm-error = "^1.4.2"` against a `1.4.2-beta.1` candidate.

**Practical severity note:** the actual protective gate (`just lint` / CI running `check_version_sync.py`) still correctly catches a real pin-drift scenario via `validate_crate_versions`, since that function is unchanged and still authoritative. The gap is that the *Rust test specifically* is now weaker than it appears and could give a developer false local confidence via `cargo test` alone — a real but narrower problem than the original release-blocking conflict, which is why both reviewers scored this "important," not "critical."

**Recommendation** (ruthless-boundary-qa's, which I agree with): apply the same successful remediation pattern used for RBQA-F001 — delete `atm_core_dep_version_matches_workspace_version` and rely on `validate_crate_versions`/`validate_lockfile` as the sole owner, or have the Rust test shell out to `check_version_sync.py` rather than reimplementing comparison logic independently.

## Separately noted — genuinely out of this PR's scope

**rust-qa-agent's QA-001** (important, filed for awareness, not blocking this PR): the internal Rust-to-Rust workspace dependency-pinning convention (bare `^x.y.z` caret requirements across ~20 crates) means a real prerelease workspace version isn't buildable today regardless of this fix — Cargo's own resolver rejects it before either version-sync check even runs. This is a pre-existing, repo-wide structural question, not something 449c641a introduced or should be expected to fix within this narrow chore's scope. Recommend routing to a dedicated architecture/rust-best-practices pass if a real prerelease release is ever planned.

## Minor

Both arch-qa and rust-qa-agent independently confirmed the previously-noted asymmetry (Python's grammar validation vs. Rust's base-only extraction) has no reachable failure path today, since Cargo's own manifest parser already rejects grammar-invalid version strings before either check runs — confirmed non-blocking, optional polish only.

## What passed clean

Full `cargo fmt --check`, `clippy -D warnings`, `python3 -m pytest .just/tests/test_check_version_sync.py` (14 passed), `cargo test -p agent-team-mail --test version_alignment` (2 passed), and `just lint version` all green at the current repo state.

## Recommendation

Apply the same fix pattern used for RBQA-F001 to RBQA-F002: delete the duplicate Rust dependency-pin test rather than patching its comparison semantics independently. This is a small, well-scoped follow-up given the template for the correct fix already exists in this same PR.

randlee:
# QA-HERMES-VERSION-ALIGN-3 — Recheck for PR #897

**Branch:** `chore/hermes-atm-version-alignment` @ `9ae109fb`
**Verdict:** **PASS**
**Deliverable completion:** 3/3 (100%)

```json
{
  "sprint": "hermes-rc-version-alignment",
  "task": "QA-HERMES-VERSION-ALIGN-3",
  "branch": "chore/hermes-atm-version-alignment",
  "commit": "9ae109fb",
  "pr": 897,
  "verdict": "PASS",
  "deliverables": { "complete": 3, "total": 3, "percent": 100 },
  "findings": { "blocking": 0, "important": 0, "minor": 0 },
  "merge_readiness": "ready",
  "merge_reason": "RBQA-F002 (the last outstanding item across 3 QA rounds) is fully resolved: the duplicate Rust test is genuinely deleted (workspace-wide search, not just single-file check), the sole canonical Python gate is confirmed wired into CI, and the new regression test is confirmed non-tautological by diffing against the prior test file"
}
```

## Reviewers (3/3 returned, unanimous PASS)

| Reviewer | Verdict |
|---|---|
| ruthless-boundary-qa | PASS — traced through what would happen under the old (deleted) looser comparison to confirm the new test is a real guard, not a tautology |
| req-qa | PASS — 100% deliverable completion, workspace-wide search (not single-file) confirms zero remaining references anywhere in the tree |
| rust-qa-agent | PASS — full `fmt`/`clippy -D warnings`/`cargo test --workspace` all green; independently confirmed PR #897 is `OPEN`/`MERGEABLE` with all 9 CI checks green |

## What was fixed and how each was verified

`crates/atm/tests/version_alignment.rs` (74 lines, including the stripped-base `release_version_base()` comparator that caused the RBQA-F002 gap) is genuinely and completely deleted — all 3 reviewers ran workspace-wide searches (not just checking the one file) and found zero remaining references, no dangling `[[test]]` declaration, and no other file reintroducing a parallel check.

`check_version_sync.py`'s `validate_crate_versions`/`validate_lockfile` are confirmed **unchanged** (byte-identical between the prior and current commit per rust-qa-agent's diff) and remain the sole owner — still doing exact-string comparison, which was always the correct behavior since crates.io publish requires an exact version match. All 3 reviewers independently traced the wiring: `.just/run_lint.py` registers `"version"` as a lint task → `justfile`'s `lint` recipe → `.github/workflows/ci.yml`'s `just lint` step. No parallel or competing gate exists anywhere in the tree.

The new regression test, `test_validate_crate_versions_requires_exact_prerelease_path_dep_pin`, was specifically checked for being a genuine guard rather than a tautology by all 3 reviewers, each tracing what would happen under the old (deleted) stripped-base comparison: with a prerelease workspace version and a stale un-bumped dependency pin, the old comparator would have reduced both to the same numeric base and passed silently; the current exact-match comparator correctly raises `SystemExit`. rust-qa-agent additionally confirmed this test scenario was structurally inexpressible before this fix (the `write_repo()` fixture gained a new `workspace_version` parameter that didn't exist previously).

## Cycle summary

This closes the full QA-HERMES-VERSION-ALIGN-1 → 2 → 3 cycle for PR #897. The critical finding from round 1 (a Rust test duplicating and conflicting with the canonical Python version-sync check) and its round-2 incomplete fix (a second Rust test given loosened, divergent semantics instead of being removed) are both now genuinely resolved by consolidating to a single canonical owner. PR #897 is clean from a QA standpoint.

PR #898, opened 2026-08-16
test(hermes-atm): multi-target wheel and CPython 3.11-3.14 compatibility

randlee:
## Summary

Requested directly by Rand after RC verification confirmed the live 1.4.2 wheel was not installed on a test profile -- adds explicit multi-Python compatibility coverage so this class of issue is caught by CI going forward.

- `hermes-atm` now declares `CPython >=3.11,<3.15` support explicitly.
- CI adds isolated build/install/contract coverage for Python 3.11, 3.12, 3.13, and 3.14.
- Test runner asserts the hermes-atm artifact is a universal `py3-none-any` wheel.

Split out from PR #897 (version-alignment fix), which it was briefly and incorrectly bundled into mid-QA-round.

Commit: `5b3d70cc`

## Test plan
- [x] Local isolated wheel proofs passed on 3.11 and 3.14: matching atm-graft wheels installed, all 17 hermes-atm tests passed each run
- [ ] Full CI matrix (3.11/3.12/3.13/3.14)
- [ ] quality-mgr QA pass

PR #906, opened 2026-08-16
fix(hermes-atm): PyPI publish-readiness CI/pipeline gaps

randlee:
Addresses 4 of the 7 blockers found by quality-mgr's QA-RELEASE-READINESS-PYPI-1 investigation (cross-cutting release-readiness review, not a sprint QA round):

1. No manylinux/musllinux wheel pipeline -- plain `maturin build` on ubuntu-latest with no `--manylinux`/auditwheel repair, wheel tag not PyPI-compliant.
2. No wheel artifact ever persisted -- CI jobs built into a tempdir that got deleted. **In progress @ 663b1e7f: retains tested graft/hermes-atm wheels outside the tempdir and uploads per-Python CI artifacts.**
3. No Windows/aarch64 wheel targets, no abi3 -- matrix expansion in progress.
4. sdist fallback unverified in CI/release automation -- in progress.

Draft PR opened immediately after first push per team policy -- work in progress, not yet ready for review.

Co-owned effort: items 5-9 (manifest registration, pyproject metadata, SkillRX redaction) are being handled in parallel by the m4-side team on a separate branch.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -283,6 +283,7 @@ jobs:
             artifact: linux-x86_64
             target: x86_64-unknown-linux-gnu
             manylinux: "2014"
+            platform_tag: manylinux_2_17_x86_64.manylinux2014_x86_64
             interpreter: python3.11
             container: ""
             smoke: native
@@ -291,6 +292,7 @@ jobs:
             artifact: linux-musl-x86_64
             target: x86_64-unknown-linux-musl
             manylinux: musllinux_1_2
+            platform_tag: musllinux_1_2_x86_64
             interpreter: python3.11
             container: ""
             smoke: musl
@@ -302,6 +304,7 @@ jobs:
             artifact: linux-aarch64
             target: ""
             manylinux: "2014"
+            platform_tag: manylinux_2_17_aarch64.manylinux2014_aarch64
             interpreter: python3.11
             container: quay.io/pypa/manylinux2014_aarch64:latest
             smoke: native
@@ -310,6 +313,7 @@ jobs:
             artifact: macos-aarch64
             target: aarch64-apple-darwin
             manylinux: "off"
+            platform_tag: macosx_11_0_arm64
             interpreter: python3.11
             container: ""
             smoke: native
@@ -318,6 +322,7 @@ jobs:
             artifact: windows-x86_64
             target: x86_64-pc-windows-msvc
             manylinux: "off"
+            platform_tag: win_amd64
             interpreter: python
             container: ""
             smoke: native
@@ -355,17 +360,10 @@ jobs:
         run: python -m pip wheel --no-deps --wheel-dir dist crates/hermes-atm
 
       - name: Verify abi3 artifact metadata
-        shell: python
         run: |
-          from pathlib import Path
-          import zipfile
-
-          wheels = sorted(Path("dist").glob("atm_graft*.whl"))
-          assert len(wheels) == 1, wheels
-          assert "cp311-abi3" in wheels[0].name, wheels[0]
-          with zipfile.ZipFile(wheels[0]) as archive:
-              wheel_metadata = next(name for name in archive.namelist() if name.endswith(".dist-info/WHEEL"))
-              assert "Tag: cp311-abi3" in archive.read(wheel_metadata).decode("utf-8")
+          python .just/verify_atm_graft_wheel.py \
+            --wheel dist/atm_graft*.whl \
+            --platform-tag ${{ matrix.platform_tag }}
 
       - name: Smoke-test native wheel on its host target
         if: ${{ matrix.smoke == 'native' }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -276,8 +276,8 @@ jobs:
           if len(wheels) != 1:
               raise RuntimeError(f"expected one downloaded atm-graft wheel, found {wheels}")
           with Path(os.environ["GITHUB_ENV"]).open("a", encoding="utf-8") as environment:
-              environment.write(f"ATM_GRAFT_WHEEL={wheels[0].resolve()}\\n")
-              environment.write(f"ATM_WHEEL_OUTPUT_DIR={Path('bridge-wheels').resolve()}\\n")
+              environment.write(f"ATM_GRAFT_WHEEL={wheels[0].resolve()}\n")
+              environment.write(f"ATM_WHEEL_OUTPUT_DIR={Path('bridge-wheels').resolve()}\n")
 
       - name: Run Hermes graft steer boundary tests
         if: ${{ needs.ci-scope.outputs.runtime == 'true' }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -279,37 +279,43 @@ jobs:
             artifact: linux-x86_64
             target: x86_64-unknown-linux-gnu
             manylinux: "2014"
-            cflags: ""
+            interpreter: python3.11
+            container: ""
             smoke: true
           - name: musllinux x86_64
             os: ubuntu-latest
             artifact: linux-musl-x86_64
             target: x86_64-unknown-linux-musl
             manylinux: musllinux_1_2
-            cflags: ""
+            interpreter: python3.11
+            container: ""
             smoke: false
           - name: manylinux aarch64
-            os: ubuntu-latest
+            # Build natively on ARM. The x86_64 cross image does not compile
+            # ring's pregenerated assembly correctly, whereas this official
+            # manylinux image provides the required PyPI ABI on ARM.
+            os: ubuntu-24.04-arm
             artifact: linux-aarch64
-            target: aarch64-unknown-linux-gnu
+            target: ""
             manylinux: "2014"
-            # ring's pregenerated aarch64 assembly requires this macro when
-            # compiled by the manylinux cross toolchain.
-            cflags: -march=armv8-a
+            interpreter: python3.11
+            container: quay.io/pypa/manylinux2014_aarch64:latest
             smoke: false
           - name: macOS aarch64
             os: macos-14
             artifact: macos-aarch64
             target: aarch64-apple-darwin
             manylinux: "off"
-            cflags: ""
+            interpreter: python3.11
+            container: ""
             smoke: true
           - name: Windows x86_64
             os: windows-latest
             artifact: windows-x86_64
             target: x86_64-pc-windows-msvc
             manylinux: "off"
-            cflags: ""
+            interpreter: python
+            container: ""
             smoke: true
     runs-on: ${{ matrix.os }}
     steps:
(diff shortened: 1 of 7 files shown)


randlee:
# QA-HERMES-PYPI-INFRA-1 — QA-1 for PR #906 (fix/hermes-atm-pypi-release-infra)

**Branch:** `fix/hermes-atm-pypi-release-infra` @ `64f3e8c69` (target: `develop`)
**Verdict:** **FAIL** (but the 4 originally-scoped blockers are genuinely closed — see below)
**Reviewers returned:** 5/5

```json
{
  "sprint": "hermes-atm-pypi-release-infra",
  "task": "QA-HERMES-PYPI-INFRA-1",
  "branch": "fix/hermes-atm-pypi-release-infra",
  "commit": "64f3e8c69",
  "pr": 906,
  "verdict": "FAIL",
  "findings": { "blocking": 0, "important": 4, "minor": 5 },
  "next_action": "Add smoke coverage for musllinux/aarch64, wire the ATM_GRAFT_WHEEL reuse seam into the test job (or delete it), add a durable platform-tag CI assertion, and investigate whether d4541d5ba's fix is masking a different root cause",
  "owner": "arch-ctm",
  "merge_readiness": "not ready",
  "merge_reason": "4 corroborated important findings remain, though the 4 originally-scoped blockers are genuinely resolved"
}
```

## The 4 originally-scoped blockers ARE genuinely closed

I independently confirmed this directly from raw CI logs (not just job pass/fail), and all 5 reviewers corroborated: the manylinux x86_64/aarch64 wheels carry genuine `manylinux_2_17`/`manylinux2014` tags, the musllinux wheel carries a genuine `musllinux_1_2` tag, wheel artifacts are persisted with `if-no-files-found: error`, Windows/macOS/aarch64 targets are all built, and the sdist path installs into a genuinely fresh `python -m venv`. The fix went further than scoped: it also added a proper `abi3` feature (one wheel per platform now covers all of Python 3.11+), which I confirmed by inspecting actual wheel filenames in CI logs (`cp311-abi3-manylinux_2_17_x86_64.manylinux2014_x86_64`, etc.) and rust-qa-agent independently confirmed by rebuilding the wheel locally.

## Important — must address before treating this as release-ready

1. **`d4541d5ba`'s fix may be masking an unidentified root cause** (rust-qa-agent). It adds `components: clippy, rustfmt` to the `hermes-atm-sdist` job's toolchain-install step — but that job never runs `cargo fmt` or `cargo clippy`, only `maturin sdist`. rust-qa-agent pulled PR #898's actual prior fix commit (`90492bf1c`) and found it applied the *identical, undocumented* remedy to a *different* job with the same characteristic (only ran `maturin build`, never used clippy/rustfmt). Two independent applications of the same unexplained fix to jobs that don't need the thing being fixed is evidence this may be a copy-paste/superstition fix over a different transient issue (e.g. a toolchain-resolution flake), not a genuine root-cause fix. I independently confirmed the before/after diff does show the line was genuinely missing — but that doesn't establish *why* adding it resolved the failure. Recommend pulling the actual failing-run logs that motivated `d4541d5ba` to confirm the real error, and documenting the causal link if it's genuine.

2. **musllinux x86_64 and manylinux aarch64 (the two newest, riskiest build targets) have zero install-time smoke verification** (req-qa's ATM-QA-002/003, corroborated by ruthless-boundary-qa). Both have `matrix.smoke: false` — no job anywhere installs and imports either wheel. The manylinux aarch64 leg runs on a native ARM64 runner that's fully capable of doing this (Python is already installed on it). musllinux would need an Alpine container step.

3. **No durable CI assertion of the platform-tag portion of the wheel filename** (req-qa's ATM-QA-001). The "Verify abi3 artifact metadata" step only checks the ABI-tag substring (`cp311-abi3`); it never inspects the platform component (`manylinux_2_17_x86_64`, etc.). My own confirmation that the tags are currently correct came from a one-time manual log inspection, not something CI enforces going forward — a future maturin/config regression producing a non-compliant tag would still pass CI green.

4. **The new wheel-reuse mechanism is both unwired and untested** (ruthless-boundary-qa's RBQA-F001, corroborated and sharpened by rust-qa-agent's QA-002). The `test` job builds `atm-graft` via a separate, differently-flagged (debug, no `--release`/`--compatibility pypi`) recipe rather than consuming the actual release-candidate wheel, even though the script already has an `ATM_GRAFT_WHEEL`/`ATM_WHEEL_OUTPUT_DIR` override seam for exactly this reuse — it's just never wired into `ci.yml` (confirmed: zero references to either env var anywhere in the workflow). rust-qa-agent additionally found this override mechanism, including its `RuntimeError` error path, is never exercised by any unit test — dead code as shipped. Note: this is not "the release wheel is untested" — the separate `hermes-atm-abi3-compat` job *does* download and test the real release wheel across all 4 Python versions — it's a duplicate-build-recipe/DRY and untested-code-path concern, not a coverage gap in the release artifact itself.

## Minor

- Untested inline Python heredoc for the `atm_graft` abi3-tag check, unlike the tested-helper convention already used for the sibling `hermes_atm` universal-wheel check (ruthless-boundary-qa's RBQA-F002).
- The `pydantic>=2,<3` constraint is hardcoded twice more in `ci.yml`'s smoke-test steps instead of reusing the existing manifest-driven `project_dependency_requirement()` helper (ruthless-boundary-qa's RBQA-F003).
- The new CI-wiring contract test does raw-text substring matching against `ci.yml` rather than parsing YAML — wouldn't catch a matched string inside a disabled block (req-qa's ATM-QA-005).
- The three new wheel/sdist/abi3-compat jobs only depend on `ci-scope`, not on the cheap `clippy`/`test` jobs completing first — a broken PR still burns multi-platform wheel-build minutes before a fast-failing lint/test job would have caught it (req-qa's ATM-QA-006, resource-efficiency only).
- A pre-existing, inert `/tmp/venv` literal in a test fixture (rust-qa-agent's QA-003, same nit flagged in earlier rounds — harmless, never touches real filesystem).

## Backlog (pre-existing, unrelated to this PR's actual diff — this PR only touches CI config and Cargo/pyproject feature flags)

rust-best-practices-agent found 2 legitimate code-quality items in `lib.rs`/`query.rs` (redundant re-validation of already-parsed typed fields; duplicated timestamp-parsing logic across two call sites) — real, but not introduced by this PR and not blocking this infra fix.

## What verified cleanly

- arch-qa: PASS, 0 architectural findings — RULE-001 through RULE-013 all compliant, abi3 feature addition confirmed purely additive with no boundary/dependency changes.
- All 6 iterative fix commits reviewed for shortcuts (by me directly via `git show`, and independently by rust-qa-agent): no `continue-on-error`, no skipped/disabled tests, no watered-down assertions anywhere. 5 of 6 are well-explained genuine fixes with clear comments (CFLAGS for ARM assembly → native ARM runner switch, container output isolation); only `d4541d5ba` lacks that explanation (see finding 1 above).
- Full workspace `fmt`/`clippy -D warnings`/`cargo test` (47 binaries) plus the Python bridge-test suite (17 tests) all pass, confirmed by rust-qa-agent both via CI logs and an independent local rebuild.
- Version discipline (workspace/`atm-graft-python`/`hermes-atm` all at `1.4.2`, dependency range correct) is mechanically enforced by the pre-existing `check_version_sync.py` machinery and unaffected by this diff.
- Items 5-9 from the original investigation (manifest registration, metadata, SkillRX redaction, versioning ADR, PyPI account setup) are confirmed correctly out of scope — untouched by this diff, as the task specified.

## Recommendation

The core infrastructure achievement is real and should not be discounted — genuine manylinux/musllinux compliance, real persisted artifacts, and abi3 support beyond what was even asked for. But 4 corroborated important findings, including one (`d4541d5ba`) that questions whether a "fix" is masking a different bug, warrant a fix round before treating this as release-ready.

randlee:
# QA-HERMES-PYPI-INFRA — Final Verdict for PR #906

**Branch:** `fix/hermes-atm-pypi-release-infra` @ `22912bcce` (target: `develop`)
**Verdict:** **PASS** (applying the production-readiness bar: does this make the artifact broken, unsafe, or rejected by the target registry — not zero-findings-of-any-severity)
**CI:** fully green, all 19 checks (independently confirmed by team-lead)

## What this PR delivers

Closes the 4 originally-scoped blockers from `QA-RELEASE-READINESS-PYPI-1`: no manylinux/musllinux pipeline, no persisted wheel artifact, no Windows/aarch64 targets, unverified sdist path — plus `abi3` support beyond scope (one wheel per platform now covers all of Python 3.11+). I independently confirmed genuine `manylinux2014`/`musllinux_1_2` tags directly from CI logs, not just job pass/fail.

## Fix cycle for this round

`QA-HERMES-PYPI-INFRA-1`'s QA-1 pass found 4 important findings on top of the already-closed blockers. All 4 are now genuinely fixed, each independently verified by me and/or an independent reviewer:

1. **Sdist toolchain rationale** — the `clippy, rustfmt` component provisioning on jobs that don't use those tools is now documented: maturin's internal nested toolchain resolution can race the hosted runner's preinstalled `cargo-fmt`, causing an `rustfmt-preview` conflict without it. Confirmed genuine, not superstition.
2. **musllinux/aarch64 smoke coverage** — all 5 wheel-build legs now get real install-time verification (`smoke: native|musl`), including a genuine `docker run` against `python:3.11-alpine` for musllinux.
3. **Platform-tag assertion durability** — a new tested standalone script (`.just/verify_atm_graft_wheel.py`) checks both the exact filename suffix and every `WHEEL` metadata tag component per matrix leg, replacing an untested inline heredoc.
4. **Release-wheel reuse wiring** — the `test` job now downloads and consumes the actual release-candidate wheel via `ATM_GRAFT_WHEEL`/`ATM_WHEEL_OUTPUT_DIR`, instead of building a separate debug-flavor copy.

## A real bug was caught and fixed mid-recheck

During the scoped recheck of finding 4, req-qa found a genuine functional bug in the initial fix: the `GITHUB_ENV`-writing f-strings used a double-escaped `\\n` instead of a real newline (`\n`), which I confirmed via direct Python execution would have corrupted the env-var wiring and broken the `test` job on every OS. This had not yet manifested in CI (the `test` job was still waiting on its new `hermes-atm-release-wheels` dependency). Flagged immediately; arch-ctm fixed it within minutes (`22912bcce`) and added a regression test that extracts and directly executes the real embedded YAML script against a temp file — genuinely proven to be a real regression guard (traced what would happen if the bug recurred: `splitlines()` would yield 1 element instead of 2, failing the assertion). Both the fix and the new test were independently re-verified by req-qa's final narrow recheck.

## Deferred to backlog (per the new production-readiness bar, not blockers)

- 5 minor findings from the original QA-1 pass (hardcoded `pydantic` constraint duplication, raw-text CI contract test instead of parsed YAML, wheel-build jobs not waiting on cheap lint/test jobs first, one inert pre-existing `/tmp` literal) — hygiene, not artifact-breaking.
- 2 pre-existing code-quality items in `lib.rs`/`query.rs` unrelated to this PR's actual diff.
- Items 5-9 from the original investigation (manifest registration, pyproject metadata, SkillRX redaction, versioning ADR, PyPI account/token setup) — explicitly out of scope, handled on a separate parallel branch.

## Recommendation

Clean for merge. The core release infrastructure is genuinely sound: real manylinux/musllinux compliance, persisted artifacts across all 5 platforms, `abi3` support, and a real functional bug was caught and fixed before it could reach a production release.

PR #911, opened 2026-08-17
feat(release): PyPI/TestPyPI publish workflow

randlee:
Builds the actual PyPI publish mechanism -- previously missing entirely (quality-mgr's original blocker #2: "no publish mechanism exists anywhere, no twine/maturin-publish step, no token, no Trusted Publisher OIDC").

- `workflow_dispatch`-only, never runs on push/PR/tag -- publishing is always a manual, deliberate action.
- Single required input selects exactly one target: `testpypi` or `pypi`. No dual-target runs.
- Reuses the existing, already-verified CI build matrix via `workflow_call` rather than duplicating the maturin-action config -- builds the same 5 native wheel pairs (manylinux x86_64/aarch64, musllinux x86_64, macOS aarch64, Windows x86_64) plus sdist.
- Fail-closed cardinality checks: exactly one wheel/sdist file per platform, 0 or 2+ fails loudly, matching the existing `if-no-files-found: error` discipline from the build jobs.
- Two separate environment-gated jobs (`environment: testpypi` / `environment: pypi`), each using its own scoped token secret (`TEST_PYPI_TOKEN` / `PYPI_TOKEN`) via Twine -- so any environment-level approval gates Rand has configured are respected, not bypassed.
- The `pypi` job only runs when the dispatch input explicitly selects `pypi`.

**Safety boundary**: arch-ctm will validate the TestPyPI path for real (trigger it, confirm packages land and install from test.pypi.org) once CI is green. The real PyPI target will not be triggered by arch-ctm under any circumstances -- that remains Rand's action alone, every time.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,7 @@
 name: CI
 
 on:
+  workflow_call:
   pull_request:
     branches: [develop, main, "integrate/*"]
   push:
--- a/.github/workflows/hermes-atm-pypi-publish.yml
+++ b/.github/workflows/hermes-atm-pypi-publish.yml
@@ -1,0 +1,123 @@
+name: Publish Hermes ATM to PyPI
+
+on:
+  workflow_dispatch:
+    inputs:
+      target:
+        description: "Package index to receive this explicitly approved release"
+        required: true
+        type: choice
+        options:
+          - testpypi
+          - pypi
+
+permissions:
+  contents: read
+
+jobs:
+  build-release-artifacts:
+    name: Build verified Hermes ATM release artifacts
+    uses: ./.github/workflows/ci.yml
+
+  prepare-publish-artifacts:
+    name: Inspect and stage PyPI artifacts
+    needs: [build-release-artifacts]
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Download manylinux x86_64 wheels
+        uses: actions/download-artifact@v4
+        with:
+          name: hermes-atm-wheels-linux-x86_64
+          path: artifacts/hermes-atm-wheels-linux-x86_64
+
+      - name: Download musllinux x86_64 wheels
+        uses: actions/download-artifact@v4
+        with:
+          name: hermes-atm-wheels-linux-musl-x86_64
+          path: artifacts/hermes-atm-wheels-linux-musl-x86_64
+
+      - name: Download manylinux aarch64 wheels
+        uses: actions/download-artifact@v4
+        with:
+          name: hermes-atm-wheels-linux-aarch64
+          path: artifacts/hermes-atm-wheels-linux-aarch64
+
+      - name: Download macOS aarch64 wheels
+        uses: actions/download-artifact@v4
+        with:
+          name: hermes-atm-wheels-macos-aarch64
+          path: artifacts/hermes-atm-wheels-macos-aarch64
+
+      - name: Download Windows x86_64 wheels
+        uses: actions/download-artifact@v4
+        with:
+          name: hermes-atm-wheels-windows-x86_64
+          path: artifacts/hermes-atm-wheels-windows-x86_64
+
+      - name: Download atm-graft source distribution
+        uses: actions/download-artifact@v4
+        with:
+          name: atm-graft-sdist
+          path: artifacts/atm-graft-sdist
+
+      - name: Verify artifact cardinality and stage upload set
+        run: python scripts/prepare_hermes_atm_publish_artifacts.py --artifacts artifacts --output publish-dist
+
+      - name: Upload verified publish set
+        uses: actions/upload-artifact@v4
+        with:
+          name: hermes-atm-pypi-publish-set
+          path: publish-dist/*
+          if-no-files-found: error
+
+  publish-testpypi:
+    name: Publish Hermes ATM to TestPyPI
+    needs: [prepare-publish-artifacts]
+    if: ${{ inputs.target == 'testpypi' }}
+    runs-on: ubuntu-latest
+    environment: testpypi
+    steps:
+      - name: Download verified publish set
+        uses: actions/download-artifact@v4
+        with:
+          name: hermes-atm-pypi-publish-set
+          path: publish-dist
+
+      - name: Install Twine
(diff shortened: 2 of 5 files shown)


randlee:
# QA-PYPI-PUBLISH-WORKFLOW-1 — QA-1 for PR #911 (feat/hermes-atm-pypi-publish-workflow)

**Branch:** `feat/hermes-atm-pypi-publish-workflow` @ `52ccb9cea` (target: `develop`)
**Verdict:** **PASS**
**CI:** green, 20/20 checks
**Reviewers returned:** 3/3 (req-qa, arch-qa, ruthless-boundary-qa)

```json
{
  "sprint": "hermes-atm-pypi-publish-workflow",
  "task": "QA-PYPI-PUBLISH-WORKFLOW-1",
  "branch": "feat/hermes-atm-pypi-publish-workflow",
  "commit": "52ccb9cea",
  "pr": 911,
  "verdict": "PASS",
  "deliverables": { "complete": 9, "total": 9, "percent": 100 },
  "findings": { "blocking": 0, "important": 0, "minor": 6 },
  "next_action": "Merge. File the 6 minor findings as fast-follow items (not blocking).",
  "owner": "arch-ctm",
  "merge_readiness": "ready",
  "merge_reason": "No blocking or important findings from any of 3 reviewers; all 7 rigorously-specified requirements independently traced and confirmed against the actual YAML/Python by every reviewer"
}
```

## All 7 requirements independently verified by all 3 reviewers, directly against the workflow YAML/scripts (not the description)

1. **workflow_dispatch-only trigger** — confirmed: `on:` block in `hermes-atm-pypi-publish.yml` contains only `workflow_dispatch:`, no push/PR/tag/schedule anywhere in the file.
2. **Single required input, no silent default** — `target: choice [testpypi, pypi]`, `required: true`, no `default:` key present anywhere. A missing/blank dispatch input causes **both** downstream `if:` conditions to evaluate false — fail-closed (neither job runs), not fail-open to a real publish.
3. **Reuses CI build matrix via `workflow_call`** — `ci.yml:4` declares `workflow_call:` as a real top-level trigger (grep confirmed exactly one occurrence in the file, not a comment); `build-release-artifacts` invokes it via `uses: ./.github/workflows/ci.yml`. arch-qa additionally traced that `ci-scope`'s doc-only-skip logic can't fire on a `workflow_dispatch` context, so the full wheel/sdist matrix always runs — no path to a silently-empty publish set.
4. **Fail-closed cardinality checks** — `exactly_one()` in `prepare_hermes_atm_publish_artifacts.py` uses `len(matches) != 1`, genuinely rejecting both 0-file and 2+-file cases. Both failure modes are exercised by tests, though only via the sdist/native-wheel call sites (see minor findings below).
5. **Two environment-gated jobs with correct equals-check conditions** — `publish-testpypi`/`publish-pypi` are separate jobs, each with a job-level `environment:` field and `if: inputs.target == '...'` (a direct equals check, not an inverted "unless" condition) gating access to correspondingly-scoped `TEST_PYPI_TOKEN`/`PYPI_TOKEN` secrets. An unexpected input value causes neither job to run.
6. **No secret leakage** — `TWINE_PASSWORD` sourced from secrets via job-level `env:`, never interpolated into a `run:` command string or CLI argument.
7. **Regression test is genuine, not tautological** — all 3 reviewers independently traced the counterfactual: `"workflow_call:"` occurs exactly once in `ci.yml`'s raw text (the real trigger), so the new test would genuinely fail if that trigger were deleted.

req-qa additionally confirmed the new test files are correctly collected by `.just/run_pytests.py` (both define proper `unittest.TestCase` subclasses — avoiding the exact silent-non-collection defect from the historical `AJ7-ATM-QA-001-TEST-NOT-COLLECTED` finding), and confirmed this PR introduces no independent version-bump logic (consistent with the ADR-049 scope boundary).

## Minor findings (not blocking — recommend as fast-follow, not required before merge)

1. **Wheel-side cardinality 2+-match branch under-tested** (ruthless-boundary-qa RBQA-F002 + req-qa ATM-QA-001, corroborating) — the shared `exactly_one()` is called at 3 sites (native wheel, universal wheel, sdist); only the sdist call site is driven through its 2+-match failure branch by a dedicated test. Recommend a test that duplicates a wheel file within one platform directory.
2. **Duplicate cardinality-check logic** (ruthless-boundary-qa RBQA-F001) — `ci.yml`'s pre-existing "Select host release atm-graft wheel" step has its own independent inline Python re-implementation of the identical "exactly one file matching this glob" decision. Recommend extracting `exactly_one()` into a shared module both consume.
3. **Regression test for `workflow_call` is a raw substring match, not YAML-structural** (ruthless-boundary-qa RBQA-F003) — genuinely fails if the trigger line is deleted, but would still pass if it were merely commented out. Recommend parsing with `yaml.safe_load` instead.
4. **No regression guard against a future accidental `default:` key** on the `target` input (req-qa ATM-QA-002) — no live defect today (confirmed no default present), but nothing stops a future edit from silently adding `default: pypi`.
5. **Unpinned `twine` install on the credentialed publish path** (req-qa ATM-QA-003) — `python -m pip install twine` has no version pin, a supply-chain risk surface specifically on a code path authenticating to a public registry with a long-lived token. Recommend pinning.
6. **GitHub Environment protection rules can't be verified statically** (ruthless-boundary-qa, notes) — whether `testpypi`/`pypi` environments have required-reviewer protection configured is repo-settings state, not visible in any file. Recommend confirming server-side before the first live `pypi`-target dispatch. Also noted: this PR uses classic token/twine auth rather than PyPI Trusted Publisher OIDC — a legitimate, least-privilege-consistent design choice (no `id-token: write` requested), but worth an explicit confirmation this was deliberate given the original blocker language mentioned OIDC as an option.

## Recommendation

Clean PASS. This is the actual PyPI publish mechanism (previously entirely missing) and all of the high-stakes safety properties I asked reviewers to verify rigorously — manual-only trigger, single-target fail-closed input, environment-scoped secrets with correct (non-inverted) gating, no secret leakage, and a genuine regression guard on the CI-matrix reuse — check out under independent, evidence-cited verification from all 3 reviewers. Recommend merge; file the 6 minor items as fast-follow hardening, and confirm GitHub Environment protection rules server-side before arch-ctm's planned manual TestPyPI dry-run.

PR #1096, opened 2026-08-29
feat(ci): prerelease-archive workflow — pre-release binaries from any ref (AR1.1)

randlee:
## Summary

Sprint AR1.1: builds manifest-declared release archives (linux/macOS/windows)
from any ref, with CI provenance, without tagging or publishing anything —
for Phase AR integration testing (docker-testbed, Mac daemon-switch parity).

- `.github/workflows/prerelease-archive.yml` (new): `workflow_dispatch` with
  `ref` (default `develop`) and optional `targets` inputs. `plan` job
  resolves the exact SHA, computes `<workspace version>-pre.<9-char short
  sha>` (e.g. `1.4.5-pre.eab52792e`), asserts it is never tag-style, and
  resolves the requested release-target matrix (default: all four manifest
  targets — macOS included, not opt-in; pass `targets=...` to narrow).
  `build` job builds and packages one archive per target. `checksums` job
  writes `checksums.txt` (`sha256sum`-compatible, `shasum -a 256 -c`
  verifiable) and `provenance.json`
  (`{version, atm_core_sha, ref, run_id, targets:[...]}`). No tagging, no
  GitHub Release, no secrets; `permissions: contents: read`; concurrency
  keyed on `ref`; `timeout-minutes` on every job.
- `.github/scripts/tests/test_prerelease_archive_workflow.py` (new): proves
  the packaging step is byte-identical to `release.yml`'s own packaging
  step (modulo the version-source expression), executes it against a
  fixture tree (tar.gz/zip layout, Windows `.exe` suffix), and asserts the
  workflow never tags/publishes/reads secrets, defaults to every manifest
  target, and rejects a tag-style version.
- `docs/plans/phase-ar/sprint-AR1-1-prerelease-archive.md` (new) and a new
  operator section in `docs/release-preflight-checklist.md`.

### Scope change from the original brief, and why

The brief asked for a new `package-archive` subcommand in
`.github/scripts/release_artifacts.py` and for `release.yml`'s inline
packaging step to be replaced with a call to it. Both files turned out to be
vendored sc-publish kit files, pinned via `release/sc-publish-pin.toml` and
installed byte-for-byte per `README.sc-publish.md` ("copied files are never
hand-edited... local drift is a defect, not a customization mechanism"),
confirmed by `test_release_artifacts.py`'s own anti-leakage inventory test
listing both files. Editing them locally here would fork the pinned kit.

So `release.yml` and `release_artifacts.py` are **unmodified** by this PR.
`prerelease-archive.yml` is an atm-core-owned addition instead (the kit's own
anti-leakage test explicitly allows consumers to add their own workflows),
and its packaging step **duplicates** — rather than edits —
`release.yml`'s "Package manifest-declared release archive" step verbatim.
A new test proves that duplication hasn't drifted. Consolidating the two
packaging steps behind one shared subcommand is the better long-term shape,
but that belongs in a PR against `randlee/sc-publish`, reviewed jointly, then
adopted here by advancing the pin — recommended as a follow-up, not done
here. Full rationale in the sprint doc.

`release.yml`'s packaging behavior itself is untouched and therefore
trivially behaviour-preserving (it was never edited).

## How to dispatch

```bash
gh workflow run prerelease-archive.yml -f ref=<sha-or-branch> [-f targets=<comma-separated-triples>]
gh run download <run-id> -n x86_64-unknown-linux-gnu   # one archive per requested target
gh run download <run-id> -n checksums                  # checksums.txt + provenance.json
```

Not dispatched and CI not polled as part of this PR, per the sprint brief.

## Test plan

- [x] `pytest .github/scripts/tests/test_prerelease_archive_workflow.py -q` — 8 passed
- [x] `pytest .github/scripts/tests/test_release_artifacts.py -q` — 58 passed, 7 skipped, 5 failed; identical failure set on `origin/develop @ eab52792e` before this branch's changes (pre-existing, unrelated: line-ceiling drift, anti-leakage inventory, a homebrew Ruby-formula assertion, two channel-recovery assertions) — this PR introduces zero new failures and fixes none of the pre-existing ones (fixing them means editing pinned kit files, out of scope here)
- [x] `just lint` — 32/32 checks passed


randlee:
## Final Quality Report — PR #1096 `feature/ar1-prerelease-archive`

**Reviewed commit: `db3488fde`** (the PR's actual GitHub head at review time — confirmed via `gh pr view 1096 --json headRefOid`). Task: AR1.1-prerelease-archive-rework, full-gate QA-1, per `docs/plans/phase-ar/sprint-AR1-1-prerelease-archive.md`.

**Verdict: FAIL — not mergeable as-is.** 3 confirmed Critical/Blocking findings, 3 Important, 3 Minor/non-blocking.

Reviewer panel (7/7 returned): req-qa, arch-qa, ruthless-boundary-qa, rust-qa-agent, rust-best-practices-agent (0 findings), rust-service-hardening-agent (skipped, no service surface), flaky-test-qa. All findings below were independently reproduced by me directly against the actual PR files/commit — not relayed on reviewer say-so.

### Blocking (Critical) — must fix before QA-2

1. **ATM-QA-001 — missing `GH_TOKEN` breaks the release-absence probe on every real run.** `.github/workflows/prerelease-archive.yml`'s "Validate prerelease tag and workspace version" step calls `gh api --include "repos/.../releases/tags/${tag}"` with no `GH_TOKEN`/`GITHUB_TOKEN` set anywhere in the step/job/workflow. Every other `gh` call in this repo's workflows sets it explicitly (verify-published-release/action.yml:37, winget-publish.yml:70, pypi-publish.yml:72, release-preflight.yml:111,141). Without it, the probe fails auth, matches neither 200 nor 404, and the job dies at "could not prove that GitHub Release is absent" — no archive ever builds. **Fix:** add `env: GH_TOKEN: ${{ github.token }}` to that step. Confirmed independently by me, req-qa, and arch-qa.

2. **FTQ-001 — new test breaks the `windows-latest` `just-lint` CI leg.** `.just/tests/test_prerelease_archive_workflow.py::test_packager_executes_windows_suffix_logic` extracts and executes the workflow's packaging heredoc verbatim, which shells out via `subprocess.check_output(["python3", ...])` (prerelease-archive.yml, literal `"python3"`, not `sys.executable`). This repo's Windows bootstrap venv only provides `.bootstrap-venv\Scripts\python.exe` (confirmed via `Justfile`'s `python_cmd` and `tools/bootstrap.py::venv_python_path()` — neither creates a `python3` alias on win32). `just lint` runs `pytests` on all 3 CI matrix OSes (`ci.yml` `just-lint` job, `.just/run_lint.py` `ALL_LINT_ORDER`), so this test fails deterministically on Windows while passing on ubuntu/macos. **Fix:** don't touch the production workflow script (would break the byte-for-byte parity test against `release.yml`) — shim a same-named `python3`(.exe) pointing at `sys.executable` onto the child subprocess's `PATH` inside the test only. Confirmed independently by me and flaky-test-qa.

3. **QA-001 — `update_lockfile()` crashes on a real, existing Cargo.lock name collision.** `.just/prerelease_tag.py::update_lockfile()` (line 129) matches `[[package]]` blocks by crate name only, with no check for a `source =` line. This repo's actual `Cargo.lock` has exactly this collision: a registry crate `sc-lint-directives` v0.5.0 (`source = "registry+..."`, lines 1955-1962) and the genuine workspace member `crates/sc-lint-directives` v1.4.5 (no `source` field, lines 1964-1970) — confirmed both by direct read of `db3488fde`'s `Cargo.lock` and by confirming `crates/sc-lint-directives` is a real `Cargo.toml` workspace member. Running the real `just prerelease-tag` today crashes with `SystemExit: Cargo.lock version for sc-lint-directives is not 1.4.5` on the very first invocation — after `update_manifest_versions()` has already written the bumped versions to disk (see QA-002), leaving a dirty, half-bumped tree. **Fix:** disambiguate blocks by absence of a `source =` line (workspace/path crates never have one) before matching by name. Confirmed independently by me and rust-qa-agent.

### Important — fix in this pass (QA-1 → QA-2 loop, no backlog deferral)

4. **QA-002** — `execute()` calls `update_manifest_versions()` (writes Cargo.toml + member manifests to disk) *before* `update_lockfile()`. When the lockfile step raises (as in QA-001, or any other lockfile inconsistency), the already-bumped manifests are left dirty on disk with no rollback. Fix: validate/lock-step the Cargo.lock update before mutating any manifest, or revert written files on failure before re-raising.
5. **QA-003** — the new test suite never drives `update_lockfile()`/`update_manifest_versions()` against a representative `Cargo.lock`; only pure-function (`patch_bump`) and text-presence assertions exist. None of the reported 9/9 or 893/11 pass counts would ever have caught QA-001. Fix: add a test using this repo's real `Cargo.lock` (or a fixture reproducing a workspace/registry name collision) driving `update_lockfile()` end-to-end.
6. **RBQA-F001** — `remote_tag_exists()` in `.just/prerelease_tag.py` and the inline tag-existence check in `release.yml`'s `gate-and-tag` job both independently reimplement `git ls-remote --exit-code --refs origin refs/tags/<tag>`, with no shared owner despite `release_artifacts.py` already centralizing equivalent predicates (`verify-version`, `verify-version-lockstep`). Fix: add a shared `release_artifacts.py` subcommand, or explicitly document in the sprint doc why the two call sites are intentionally separate (different credential context: local git vs. CI `contents: write`).

### Minor — fix in this pass (doc-only)

7. **ATM-QA-002 (req-qa)** — sprint doc's reused-subcommand list omits `build-plan`, which the workflow also calls (pre-existing/unmodified subcommand, not a boundary issue — doc completeness gap only).
8. **RBQA-F002** — `prerelease-archive.yml`'s in-file comment points to `.github/scripts/tests/test_prerelease_archive_workflow.py` (the pinned vendored-kit test tree) but the actual test lives at `.just/tests/test_prerelease_archive_workflow.py` — stale/misleading path, doc-only fix.

### Non-blocking / backlog (pre-existing, systemic — not introduced by this diff)

9. **QA-004** — `just validate` fails at the `sc-ecosystem-preflight` stage (stale sc-ecosystem/Wyvern pins, Wyvern missing from PATH). Identical failure set reproduces on develop and on every recent PR in this repo; not introduced here. No action required for this PR.

### Review-target dispositions (8 items from the task assignment)

- **Version scheme is tag+patch-bump only**: CLOSED — `workspace_version()`/`patch_bump()` enforce stable X.Y.Z, no short-SHA scheme (`test_workflow_uses_patch_bumped_stable_versions_without_short_sha_scheme` mechanically enforces this).
- **`just prerelease-tag` correctness/atomicity**: OPEN — blocked by QA-001/QA-002 (not atomic, crashes on real repo state).
- **Workflow triggers-only-on-tag, validates lockstep, no-existing-release, no-develop-reachability**: CLOSED except the no-existing-release check is currently broken by the missing `GH_TOKEN` (finding #1); develop-reachability is deliberately and correctly absent (verified: zero `merge-base`/`is-ancestor` references, mechanically enforced by `test_workflow_is_tag_only_and_validates_tag_version_and_release_absence`).
- **Packaging equivalence, no tag creation, `contents: read` retained**: CLOSED — verified genuine byte-for-byte extraction/comparison against `release.yml` (not a weak/hardcoded check), `permissions: contents: read` confirmed at workflow level, no tag-creation call in the workflow itself.
- **New test file location, `discover_repo_root` usage, 9/9 passing; independent 893/11 pytest count**: CLOSED — reproduced exactly: `.just/tests/test_prerelease_archive_workflow.py` 9/9 at `db3488fde`; `.just/run_pytests.py` 893 passed/11 skipped at `db3488fde`, matching dev's claim exactly.
- **Pre-existing `.github/scripts/tests/` failure count on develop `ae03b6a91`**: CLOSED — reproduced exactly: 7 failed/95 passed/8 skipped/3 subtests, matching dev's claim and resolving the earlier 5-vs-7 discrepancy from PR #1097. Same set reproduces unchanged on the PR branch (no new failures introduced).
- **No crates/daemon/pinned-kit files touched**: CLOSED — verified via `git diff --stat ae03b6a91...db3488fde`: exactly 6 files touched (`prerelease-archive.yml`, `.just/prerelease_tag.py`, `.just/tests/test_prerelease_archive_workflow.py`, `Justfile`, sprint doc, release-preflight checklist doc), zero touches to `crates/`, daemon, or `release/*.toml`/`release/*.json`.
- **`db3488fde` is a genuine merge, `eba6330f0` lineage intact**: CLOSED — verified 2-parent merge (`252bd33de`, `eba6330f0`), `eba6330f0` confirmed ancestor via `git merge-base --is-ancestor`.

### Process note

The shared worktree accumulated local, **unpushed** commits during this review (`42e202936` → `f0c7b2488`, authored by Rand Lee) that already implement fixes for findings #1, #2, and #7 above (GH_TOKEN env, python3-shim test fix, build-plan doc mention) — confirmed by direct diff read. **None of this is pushed to the PR** (GitHub head is still `db3488fde`); QA-001/QA-002/QA-003/RBQA-F001/RBQA-F002 are not addressed by that local work either. Recommend the branch stay frozen (no further local pushes) until the fix-assignment for all 6 fix-now items above lands as a single reviewable commit, to avoid re-litigating a moving target across QA rounds.

Full reviewer detail available on request. This report supersedes the two interim updates sent during the review.

randlee:
## Final Quality Report — QA-2 — PR #1096 `feature/ar1-prerelease-archive`

**Reviewed commit: `5bd19c7f911b25686c03a5261295a5dece461d56`** (confirmed as PR head via `gh pr view --json headRefOid`). Task: AR1.1 fix-round re-verification, narrow QA-2 panel (req-qa, arch-qa, ruthless-boundary-qa, rust-qa-agent, flaky-test-qa — rust-best-practices-agent/rust-service-hardening-agent correctly omitted per QA-2 policy).

**Verdict: FAIL — not mergeable.** Confirmed by live CI evidence plus 2 Blocking + 3 Important + 5 Minor findings, all independently verified by me directly against the real files/CI logs, not relayed on reviewer say-so.

### Blocking — CI is red, right now, on this exact commit

**`Just lint (windows-latest)` is FAILING at `5bd19c7f9`** (`gh pr view 1096` statusCheckRollup; run 33235094740, job 99054584365). This directly contradicts the QA-2 dispatch's claim of "full `.just/run_pytests.py` 897 passed/11 skipped ... No Rust source touched this round" — that claim was evidently not validated against real Windows CI. Pulled the actual `just-lint-logs-windows-latest` artifact:

```
FAIL: test_plan_step_exercises_authenticated_release_probe
  File ".just/tests/test_prerelease_archive_workflow.py", line 232
    self.assertEqual(result.returncode, 0, result.stderr)
AssertionError: 1 != 0 :
Ran 906 tests in 198.629s
FAILED (failures=1, skipped=37)
```

`result.stderr` is empty — the extracted bash step exits 1 with no diagnostic output, on Windows only (ubuntu-latest and macos-latest `Just lint` both pass). This is the exact test added this round to verify the GH_TOKEN fix (ATM-QA-001), and it is the third-and-different failure mode from what any of the 5 dispatched reviewers hypothesized — not the `_python3_shim`/stdlib-resolution risk flaky-test-qa flagged (FTQ-010, different test) and not the `cargo update --offline` cold-cache risk (FTQ-011, different test again, though I independently confirmed via `.github/workflows/ci.yml` that `just-lint`'s only cache is `~/.cargo/bin`/`.crates2.json`/`.bootstrap-venv` — never `~/.cargo/registry` — so FTQ-011 is a real, separate, still-live risk for `test_candidate_bump_updates_actual_lockfile_collision_safely` even though it isn't what broke this run).

**Action required before any further QA round:** root-cause the actual Windows failure (get the real stderr — the harness appears to be swallowing it; `.just/logs/20260829050434-pytests.log`'s "stderr:" section is garbled/interleaved with unrelated cargo-deny/benchmark-publish output and is not trustworthy) and fix it, **and** separately address FTQ-010 (prefer `os.symlink(sys.executable, ...)` over `shutil.copy2` for the python3 shim — copy-based shims are exactly the class of host/build-dependent failure this fix round was meant to close) and FTQ-011 (warm `~/.cargo/registry` in `just-lint`'s cache/bootstrap step, or stub `cargo update` in the new lockfile test, before it silently starts failing cold on a fresh runner image).

### Blocking — doc/plan conflicts confirmed by direct read

1. **ATM-QA-101** — `docs/plans/phase-ar/sprint-AR1-1-prerelease-archive.md:91` states "No `crates/` or daemon changes," but `python_project_paths()` (`.just/prerelease_tag.py:69-70`) now globs and bumps `crates/*/pyproject.toml` on every real run — confirmed directly (`grep` shows both the doc line and the code). The behavior itself is correct/necessary (these 3 files are registered `python_packages` enforced by `verify-version-lockstep`), so the fix is to correct the doc's Constraints wording, not the code. **Fix:** narrow the constraint text to explicitly scope out manifest-only `pyproject.toml` version bumps.
2. **ATM-QA-102** — `docs/project-plan.md:1467` (§50, "Phase AR — codex-atm Host Integration") is explicitly `[PLANNING — REQUIREMENTS DRAFT, OFF-INTEGRATION BRANCH]`, "No sprints are cut," "not merged to develop" — confirmed by direct read. Yet `sprint-AR1-1-prerelease-archive.md` is an active, cut sprint (this PR) under the same "AR1" namespace, targeting `develop`. This predates the QA-1→QA-2 fix round (not new-to-this-diff), so it should be resolved as a fast parallel doc-hygiene fix (rename the sprint out of the reserved namespace, or reconcile `project-plan.md` to acknowledge this cut release-engineering line) rather than blocking on code rework — but it is a genuine plan-baseline conflict and must be resolved before merge.

### Important — new to this diff, no backlog deferral

3. **RBQA-F002** (confirmed via `grep` — `release_artifacts.py:804` `cmd_sync_python_version`, `:768` `_python_project_declares_dynamic_version`) — `python_project_paths()`/`replace_package_version()` reimplements pyproject.toml version-bumping from scratch instead of shelling out to the already-vendored `release_artifacts.py sync-python-version` (which this same file already calls for `build-plan`/`verify-version`/`verify-version-lockstep`). The homegrown regex has no dynamic-version awareness; no active failure today (both pyproject.toml files are statically versioned) but fragile against a future maturin-dynamic manifest. **Fix:** shell out to `sync-python-version` per discovered path, mirroring the existing subprocess pattern.
4. **RBQA-F003** (confirmed via `grep` — `.just/lint-config.toml:231-236` `[version_sync.winget]`, `.just/check_version_sync.py:195` `validate_winget_manifests`) — `winget_manifest_paths()` hardcodes `.winget/*.yaml` and does a blind whole-file `str.replace` instead of reading the canonical config-driven, field-targeted definition that already exists and is enforced elsewhere in this same codebase. The two definitions currently agree by coincidence; the write side is strictly weaker than the read/validate side. **Fix:** read glob/field names from `lint-config.toml` via `load_lint_config`, and use a field-targeted rewrite matching `validate_winget_manifests()`'s own regexes.
5. **RBQA-F005** (confirmed — `git status --porcelain --ignored` in the worktree shows only `__pycache__`/`.pytest_cache`/log junk today, no active leak) — `candidate_changes()`'s `shutil.copytree(repo_root, candidate, ignore=...)` uses a hardcoded 4-item directory denylist rather than a tracked-files-only copy. Structurally unbounded: any future untracked file at repo root (a stray `.env`, credential dump) gets silently copied into a `/tmp` staging directory on every real invocation. **Fix:** build the candidate from tracked content only (`git worktree add`, `git archive HEAD | tar -x`, or `git ls-files -z`).

### Minor — new to this diff, no backlog deferral

6. **ATM-QA-002-RECHECK** — sprint doc's subcommand list still omits `cargo-build-bin-args` and `release-package-config` (confirmed real workflow calls at `prerelease-archive.yml:137,165-171`) even after QA-1's fix added `build-plan`.
7. **ATM-QA-103 / RBQA-F004** (same underlying gap) — sprint doc's Delivered bullet 2 doesn't disclose the new pyproject.toml/.winget bump scope or the new `candidate_changes()`/`validate_candidate()`/`write_and_commit()`-with-rollback architecture.
8. **RBQA-F006** — the `shutil.copytree` ignore-closure is duplicated verbatim between `prerelease_tag.py:214-219` and the test file (`test_prerelease_archive_workflow.py:349-354`) instead of one shared helper — will silently go stale if the production denylist changes (e.g., to fix RBQA-F005).
9. **ATM-QA-105** — no direct unit-test assertion on `.winget/*.yaml` or the other 2 pyproject.toml files post-bump; only `Cargo.lock` and `atm-query-python/pyproject.toml` get explicit assertions, the rest rely on an indirect, generic `verify_lockstep` `SystemExit`.
10. **QA2-001** — dev's self-report claimed "focused suite 17/17"; independently reproduced by rust-qa-agent as 13/13 (the file has exactly 13 test methods). Reporting-accuracy issue only, not a code defect — correct the self-report.

### Disposition of carried-forward QA-1 item

- **RBQA-F001** — the tag-existence-check duplication (`remote_tag_exists()` vs. `release.yml`'s `gate-and-tag` job) is still physically unchanged; only a TODO comment + filed tracking issue (`randlee/sc-publish#79`, confirmed real/open) was added. **Accepted as-is** — unifying it would require editing the pinned/vendored `release.yml`, out of this PR's authority. No further action required.

### Reproduced and confirmed clean (no action needed)

- `cargo fmt --check` / `clippy -D warnings` / `cargo test --workspace`: pass, and `git diff db3488fde 5bd19c7f9 -- '*.rs'` is empty — zero Rust touched (rust-qa-agent, independently confirmed by me via the same diff command against the real repo).
- No pinned/vendored kit files touched: confirmed via `git diff --stat db3488fde 5bd19c7f9 -- .github/scripts .github/workflows/release.yml release/ README.sc-publish.md` — empty diff.
- The real `sc-lint-directives` Cargo.lock collision (QA-001 from QA-1) is genuinely and correctly resolved — rust-qa-agent hand-verified the resulting lockfile bytes directly (0.5.0/registry unchanged, 1.4.5→1.4.6/no-source correctly bumped, all dependent edges updated), not just trusting the new test's own assertions.
- `write_and_commit()`'s rollback (QA-002 from QA-1) is sound for every code path this PR actually exercises (all touched paths are pre-existing tracked files); rust-qa-agent flagged one latent, not-yet-exercised edge case for hypothetical future newly-created-file callers — informational only, not a finding against this PR.
- `sc-ecosystem-preflight` (QA-004, carried from QA-1): reconfirmed still failing identically and systemically across develop/every recent PR (3 unchanged blockers). Non-blocking backlog, unaffected by this PR.
- `just lint` 32/32 and `just validate` (all stages pass except the systemic preflight stage above) reproduced independently by rust-qa-agent on this machine — but note this is not what CI itself is reporting on windows-latest (see the CI Blocking item above); the discrepancy between "passes locally" and "fails in real CI" is itself the headline concern of this round.
- arch-qa: PASS, 0 findings — cargo-update delegation architecturally sound and strictly safer than the string-matching it replaces; copy-validate-then-apply pattern appropriate for this blast radius; new helper functions are not unreviewed boundary reaches (winget glob matches the pre-existing lint-config entry, python glob targets the same file class the vendored kit already recognizes).

### Recommendation

Route back to `arch-ctm` via the standard fix-assignment path. Given CI is demonstrably red on the actual PR head, priority order should be: (1) root-cause and fix the real Windows `just-lint` failure with actual stderr in hand — do not re-guess from the reviewers' hypotheses alone; (2) address FTQ-010 (symlink-based shim) and FTQ-011 (registry cache) defensively since both are live, unresolved CI-reliability risks even though neither is confirmed as *this* run's specific cause; (3) the two Blocking doc/plan conflicts (can run in parallel, doc-only); (4) the 3 Important + 4 Minor boundary/doc findings. Recommend the branch stay frozen from further ad hoc local commits until this full set lands as one reviewable QA-3 candidate.

randlee:
## Final Quality Report — QA-3 — PR #1096 `feature/ar1-prerelease-archive`

**Reviewed commit: `749b1ad2e58ca1d815157eb7c548c8217d98fd4e`** (confirmed as PR head via `gh pr view --json headRefOid`, and via each reviewer's own worktree-HEAD check). Narrow QA-3 panel (req-qa, arch-qa, ruthless-boundary-qa, rust-qa-agent, flaky-test-qa), re-verifying the fix round claimed to close all 9 QA-2 items.

**Verdict: FAIL — not mergeable.** The headline QA-2 Blocking item (live Windows CI failure) is **still failing, unchanged**, and this round's fix work introduced **one new Blocking regression**. All findings below were independently reproduced by me directly against real files/live CI, not relayed on reviewer or dev say-so.

### Blocking

1. **WIN-CI-001 — NOT fixed. Live CI fails identically on `749b1ad2e`.** `Just lint (windows-latest)` (run `33236350532`, job `99057899534`) completed `FAILURE` at `2026-08-29T05:38:46Z`. I pulled the real uploaded artifact (`just-lint-logs-windows-latest` → `20260829053606-pytests.log`) myself:
   ```
   FAIL: test_plan_step_exercises_authenticated_release_probe
     File ".just\tests\test_prerelease_archive_workflow.py", line 250
       self.assertEqual(result.returncode, 0, result.stderr)
   AssertionError: 1 != 0 :
   Ran 906 tests in 153.304s
   FAILED (failures=1, skipped=37)
   stderr:
   ```
   This is the **exact same test, exact same empty-stderr symptom** as the failure reported in the QA-2 report. The dev's claimed root cause ("Windows Git Bash could not execute the extensionless POSIX stub scripts") and its fix (Python-backed `.cmd`/symlink launchers for `gh`/`jq`/`python3`) did not move this failure at all — req-qa independently confirmed the identical failure via the same artifact; rust-qa-agent and flaky-test-qa both validated only *locally* and both asserted "the Windows CI failure mechanism is fixed" without checking the real run — a repeat of the exact same overclaim pattern QA-2 caught (self-reported all-green vs. red real CI). **I am not accepting "fixed, confirmed by local run" from any source again without a green real CI run to back it.**

   New lead for the fix team, found while reading the failing test directly: `test_plan_step_exercises_authenticated_release_probe` (`.just/tests/test_prerelease_archive_workflow.py:185-250`) never installs a `python3` shim (only `_python_command_shim(tmp_path, "jq", ...)` and `_python_command_shim(tmp_path, "gh", ...)`), unlike its sibling packaging test. The extracted shell step it runs calls `python3 .github/scripts/release_artifacts.py build-plan ...` three times before ever reaching the `gh`/`jq` calls. On the real windows-latest runner, whatever `python3` this resolves to (or fails to resolve to) on bare `PATH` is untouched by this round's fix and has never been directly investigated. Recommend instrumenting this specific test with real diagnostics (`bash -x`, or teeing subprocess output to files rather than relying on captured pipes, since two rounds in a row have produced an empty `stderr`) before guessing another hypothesis.

2. **WINGET-URL-001 (new regression introduced by this round's RBQA-F003 fix)** — `sync_winget_manifests()`'s new `installer_pattern` regex (`.just/check_version_sync.py:279-289`) only advances the version in the URL's **directory** segment (`/download/vX.Y.Z/`) and leaves the **filename**-embedded version stale, because its lazy prefix group anchors on the first semver match and treats everything after it (including the second embedded version) as an untouched literal `suffix`. Confirmed **three independent ways**: I reproduced it by hand against the real `.winget/randlee.agent-team-mail.yaml` (`InstallerUrl: .../download/v1.4.5/atm_1.4.5_x86_64-pc-windows-msvc.zip`, real filename scheme confirmed at `release.yml:206`); ruthless-boundary-qa found it independently; rust-qa-agent found it independently by directly invoking `candidate_changes()` and reading the generated content, producing:
   ```
   InstallerUrl: https://github.com/randlee/atm-core/releases/download/v1.4.6/atm_1.4.5_x86_64-pc-windows-msvc.zip
   ```
   Neither the validator (`extract_version_from_url()` short-circuits on the first `/download/vX.Y.Z/` match, never inspecting the filename) nor the new test assertion (`assertIn("releases/download/v1.4.6/", winget)` — a directory-prefix substring only) catches this. Every real prerelease bump going forward would silently commit a `.winget` manifest whose installer URL points at a file that will never exist. **Fix:** replace every semver-shaped occurrence of the *old* version within the `InstallerUrl` value (e.g. `re.sub(re.escape(old_version), new_version, value)` scoped to the field), and harden `extract_version_from_url`/`validate_winget_manifests` to check *all* embedded version occurrences, not just the first, so a future regression like this fails `just lint` instead of shipping silently. Extend the test to assert the full filename segment.

### Minor

3. **(arch-qa)** Winget field-extraction regex-construction code is still duplicated line-for-line between `validate_winget_manifests()` and `sync_winget_manifests()`, even though both now correctly share one `version_sync_config()`-sourced field/glob definition (no drift risk on *values*, just unreused code). `.just/check_version_sync.py:195-248, 251-298`.
4. **(arch-qa / rust-qa-agent, same root cause as Blocking #2)** No negative-path test exists for `copy_tracked_files()`'s boundedness/symlink handling, for the `sync-python-version` subprocess hard-failure path, or for the winget installer-filename segment — coverage gaps that let WINGET-URL-001 ship undetected.
5. **(arch-qa)** `update_manifest_versions()` unconditionally reports all `python_project_paths()` as "changed" even when `sync-python-version` reports nothing to sync (dynamic-version case) — harmless no-op write, but imprecise relative to the function's contract. `.just/prerelease_tag.py:140-146`.
6. **(flaky-test-qa, reasoned but NOT confirmed as active in this run)** `_python3_shim`'s Windows branch produces only a `.cmd` file, no `.exe`. The packaging test's nested `subprocess.check_output(["python3", ...])` call (`.github/scripts` heredoc, no `shell=True`) is a native Python list-form call — `CreateProcess` on Windows does not do shell-style `PATHEXT`/`.cmd` resolution the way Git Bash does (the well-known `npm.cmd` gotcha), so this specific call could plausibly `FileNotFoundError` on Windows. **I checked the real CI log for this**: only 1 test failed this run (`test_plan_step_exercises_authenticated_release_probe`); `test_packager_executes_windows_suffix_logic` is not among the failures, so this specific mechanism did **not** manifest as a live defect in this run — flagging as a real but currently-latent risk worth a defensive fix (point the nested call at `sys.executable` instead of `"python3"`), not as a second confirmed CI blocker.
7. **(flaky-test-qa)** The new `~/.cargo/registry` cache in `just-lint` has a residual gap on the very first run ever under a given cache key (no exact-key or prefix match yet). Currently masked in practice because `.just/run_lint.py`'s `CARGO_LINT_ORDER` runs `clippy` (which warms the registry via network fetch) before the `pytests` stage that needs `cargo update --offline` — but this is an incidental ordering dependency, not a guarantee, and would break if lint task ordering/scoping ever changes.

### Confirmed genuinely fixed (independently re-verified, not just claimed)

- **ATM-QA-101** (constraints wording) — rewritten correctly, touched-file-allowlist test added.
- **ATM-QA-102** (Phase AR/AS namespace) — sprint doc moved to `docs/plans/phase-as/sprint-AS1-1-prerelease-archive.md` (confirmed a real git rename, old path gone, not duplicated), all AR/AR1 references in this PR's own files updated; `docs/project-plan.md` itself confirmed genuinely untouched (`git diff 5bd19c7f9 749b1ad2e -- docs/project-plan.md` empty) per fenix's deliberate ruling — **note this is explicitly provisional** pending fenix/Rand's final confirmation of the phase letter, so treat as conditionally-closed for this PR's purposes, not a permanently settled decision.
- **ATM-QA-002-RECHECK** (subcommand list) — all 6 real subcommands now named.
- **ATM-QA-103 / RBQA-F004** (undisclosed architecture) — Delivered section now discloses the pyproject.toml/.winget bump scope and the candidate/validate/rollback design.
- **QA2-001** (17/17 vs 13/13) — corrected to accurate 13/13.
- **RBQA-F001** — unchanged, accepted-as-is disposition still holds (TODO + `randlee/sc-publish#79` still present).
- **RBQA-F002** (pyproject reimplementation) — genuinely closed structurally: delegates to `release_artifacts.py sync-python-version` via subprocess with a hard `SystemExit` on nonzero exit, not a swallowed failure. Confirmed by arch-qa and ruthless-boundary-qa independently.
- **RBQA-F005** (copytree denylist) — genuinely closed: `copy_tracked_files()` copies exactly `git ls-files -z` output, correctly preserves symlinks via `os.readlink`/`symlink_to` rather than dereferencing. Confirmed by arch-qa and ruthless-boundary-qa independently.
- **RBQA-F006** (duplicated ignore-closure) — closed; both `candidate_changes()` and the test fixture call the same shared `copy_tracked_files()`.
- **FTQ-010** (python3 shim POSIX branch) — genuinely resolves the original stdlib-resolution concern: now `os.symlink(sys.executable, ...)`, the same technique used by pyenv/uv.
- **FTQ-011** (cold cargo registry cache) — addressed for the steady-state case (see Minor #7 for the residual first-run caveat).
- Local validation reproduced accurately this round (rust-qa-agent): zero Rust touched, `cargo fmt`/`clippy`/`cargo test` clean, `python3 .just/run_pytests.py` 897 passed/11 skipped (exact match), focused suite 13/13, `just lint` 32/32, `just validate` fails only on the pre-existing systemic `sc-ecosystem-preflight` blockers (unchanged from every prior round). **However — as with QA-2 — local-green does not mean CI-green; that gap is precisely what Blocking #1 is.**

### Recommendation

Route back to `arch-ctm`/fenix. This is now two consecutive rounds where a "root-caused and fixed" claim for the Windows CI failure was not actually verified against a real CI run before being reported as done — **do not close this item again without pasting the actual green `Just lint (windows-latest)` run URL.** Priority order: (1) instrument and actually root-cause `test_plan_step_exercises_authenticated_release_probe`'s Windows failure with real diagnostics (the missing `python3` shim in that specific test is an untried, concrete lead); (2) fix WINGET-URL-001 (all-occurrences version substitution) and harden the validator/test to catch a recurrence; (3) apply flaky-test-qa's defensive fix for the packaging test's nested `python3` call regardless of whether it's confirmed live yet; (4) the 3 Minor arch-qa/test-coverage nits. Recommend the branch stay frozen from ad hoc local commits until this set lands as one QA-4 candidate, and recommend the fix round include pasting the actual CI run URL/conclusion in the completion report this time, not just a local-test summary.

rand-lee (approved):

rand-lee (approved):

randlee:
## Final Quality Report — QA-4 — PR #1096 `feature/ar1-prerelease-archive`

**Reviewed commit: `4b93114f539abce7f17b78f74cac629e4a6ee97f`** (confirmed as real PR head via `gh api repos/randlee/atm-core/pulls/1096`, and confirmed the worktree used by every reviewer this round was genuinely checked out at this exact SHA — `git status` shows `HEAD detached at 4b93114f5`, working tree clean). Narrow QA-4 panel (req-qa, arch-qa, ruthless-boundary-qa, rust-qa-agent, flaky-test-qa), re-verifying closure of all QA-3 items.

**Verdict: FAIL — not mergeable.** Both QA-3 Blocking items are now genuinely fixed, verified independently by me and multiple reviewers, including real Windows CI evidence. But one QA-3 Minor item's "fixed" claim is false — confirmed unfixed by three independent readers — and this round's own new code introduced 3 new minor findings. Per this repo's QA-2+ merge gate (0 Blocking + 0 Important + 0 Minor, no exceptions, no backlog deferral for new-to-the-diff findings), any of these blocks merge on their own.

### Confirmed genuinely fixed (independently re-verified against real evidence, not claims)

1. **WIN-CI-001 — genuinely fixed.** I independently pulled the real check-run for the actual PR head: `Just lint (windows-latest)`, run `33237326304`, job `99060476170`, `head_sha: 4b93114f539abce7f17b78f74cac629e4a6ee97f` (confirmed via `gh api repos/randlee/atm-core/actions/runs/33237326304` — this is a genuine `pull_request`-triggered run on this exact head, not a stale citation), `conclusion: success`. I pulled the actual job log myself (`gh api .../jobs/99060476170/logs --allow-escape-sequences`) and confirmed the `run_lint.py` output shows `pytests passed [203.8s]` and `lint passed: 32 check(s) succeeded` — the previously-failing `test_plan_step_exercises_authenticated_release_probe` is inside that passing stage. Root cause and fix are sound on inspection too: the new `_bash()` resolver (`.just/tests/test_prerelease_archive_workflow.py:163-194`) explicitly walks `git --exec-path` for a real Git Bash binary and rejects any candidate under `%WINDIR%\System32` (the WSL launcher stub that was silently swallowing the test). All 5 reviewers independently read this code and found it sound; flaky-test-qa additionally confirmed no new flakiness in the resolver logic itself (deterministic candidate ordering, no timing/shared-state dependency), with one non-blocking suggestion (add a direct/mocked unit test for the resolver rather than relying only on the end-to-end Windows CI round-trip — reasonable given this is the third round on this exact bug, but not required to close this item).

2. **WINGET-URL-001 — genuinely fixed, three-way independently reproduced.** `extract_versions_from_url()` (plural, `re.findall`) now returns every embedded version occurrence; `validate_winget_manifests()` fails if any occurrence mismatches; `sync_winget_manifests()` now does a global digit-boundary-guarded substring substitution (`replace_version_occurrences()`) of the literal old version across the whole `InstallerUrl` value. Reproduced independently three ways: (a) me, by hand-tracing the regex against the real `.winget/randlee.agent-team-mail.yaml`; (b) ruthless-boundary-qa, by hand-simulating the original defect shape and confirming the new validator now catches it; (c) rust-qa-agent, by directly invoking `sync_winget_manifests()` against a fixture built from the real manifest shape and showing the actual before/after string: `.../download/v1.4.5/atm_1.4.5_....zip` → `.../download/v1.4.6/atm_1.4.6_....zip` — both segments bumped correctly. Test coverage (`test_candidate_bump_updates_actual_lockfile_collision_safely`) now asserts the full filename-embedded segment, not just the directory-prefix substring, closing the QA-3 test gap too.

3. **flaky-test-qa's FTQ-012 defensive fix** (nested nested nested `subprocess.check_output(["python3", ...])` → `sys.executable`) — genuinely fixed in both `.github/workflows/prerelease-archive.yml` and `release.yml`. req-qa additionally confirmed `release.yml`'s post-fix content is byte-identical to `develop`'s pinned kit file, so this isn't a hand-edit of vendored kit code — it's an upstream-consistent fix.

4. **QA-3 Minor (a)** — winget field-extraction duplication between `validate_winget_manifests()`/`sync_winget_manifests()` — genuinely fixed via shared `winget_settings()`/`winget_field_pattern()`/`winget_field_values()`/`replace_winget_field_values()` helpers, confirmed by arch-qa and req-qa independently.

5. **QA-3 Minor (b)** — negative-path test gaps — genuinely closed: req-qa confirmed 3 new tests (`test_copy_tracked_files_excludes_untracked_content_and_preserves_symlinks`, `test_sync_python_version_propagates_owner_failure`, and a winget filename-segment-mismatch test) are present and exercise exactly the gaps QA-3 flagged.

6. **Local validation** (rust-qa-agent): zero Rust files touched (re-confirmed against the correct `develop` merge-base after catching and correcting a stale-ref false positive from an earlier local ref); `cargo fmt --check`/`cargo clippy -D warnings`/`cargo test` all clean (1514 passed, 0 failed, 2 ignored); `python3 .just/run_pytests.py` 897 passed/11 skipped/0 failed (908 total, exact match to every prior round); `just lint` 32/32. **This is real CI-backed evidence this round, not just local claims** — Blocking #1 above is independently confirmed on the actual live windows-latest run, not only locally.

### Blocking

1. **QA-3's Minor item (c) — `update_manifest_versions()` unconditional "changed" reporting — claimed fixed, confirmed NOT fixed.** I independently traced `.just/prerelease_tag.py:116-150` myself: the `before` snapshot dict (line 121) is built only from `cargo_manifest_paths(repo_root)`; `python_project_paths(repo_root)` entries are never added to it. The final filter (`if path not in before or path.read_text(...) != before[path]`) is therefore unconditionally `True` for every `pyproject.toml`, regardless of whether `sync_python_version()` actually changed anything. **Independently confirmed by arch-qa and req-qa as well** — three-way convergent finding, not a single reviewer's interpretation. This directly contradicts the sprint doc's own deliverable text (`candidate_changes()` "returns the exact version-manifest allowlist" — `docs/plans/phase-as/sprint-AS1-1-prerelease-archive.md:51-52`). Worse, it is now **locked in by a test**: `test_candidate_bump_updates_actual_lockfile_collision_safely` (`.just/tests/test_prerelease_archive_workflow.py:389-396`) asserts unconditional inclusion of every `python_project_paths()` entry, including a dynamic-version project with no literal version field to change — i.e. the regression test codifies the bug as expected behavior rather than testing the documented contract. Escalating from QA-3's "Minor" to Blocking because (a) it was explicitly claimed fixed this round and is not, repeating this PR's now-familiar overclaim pattern, and (b) a test now actively protects the wrong behavior, meaning a future fix attempt would need to also fix the test or it will look "already covered." **Fix:** seed `before` with pre-read content for every `python_project_paths()` entry (and any other tracked-manifest family) before calling `sync_python_version()`, apply the same before/after diff filter uniformly, and update the regression test to assert only genuinely-changed files appear (including a no-op case for a dynamic-version `pyproject.toml`).

### Minor

2. **(req-qa / ruthless-boundary-qa, found independently by both)** `extract_version_from_url()` (singular) is dead code left over from the WINGET-URL-001 fix — zero callers anywhere in the repo (confirmed by repo-wide grep by both reviewers), and its own docstring's claim of "compatibility with existing callers" is false since no such callers exist. `.just/check_version_sync.py:55-59`. **Fix:** delete it.

3. **(ruthless-boundary-qa)** `replace_version_occurrences()`'s regex boundary guard is asymmetric: the leading lookbehind `(?<![0-9.])` excludes a preceding digit or dot, but the trailing lookahead `(?![0-9])` does not also exclude a following dot. Hand-traced: the pattern would partially match inside a longer dotted token (e.g. old_version `1.2.3` embedded in an unrelated `1.2.3.4`). Currently harmless only because the real manifest's `InstallerUrl` never contains such an adjacent token — not because the regex forbids it — and no test exercises this adjacency. `.just/check_version_sync.py:118`. **Fix:** change the trailing lookahead to `(?![0-9.])` for symmetry, add a direct unit test for `replace_version_occurrences()`/`extract_versions_from_url()` covering the adjacency case.

4. **(rust-qa-agent)** The new POSIX shim files (`_python3_shim`'s extension-less `python3` file, `_python_command_shim`'s extension-less `name` file, both `#!/bin/sh` scripts) are written via `Path.write_text(...)` with no `newline=` argument. On native Windows Python this triggers universal-newline translation, embedding a `\r` into the shebang line. This is exactly the class of Windows-only defect that can't be reproduced on non-Windows machines and has burned this PR twice already — it happens to work today only because current Git-Bash builds tolerate a trailing `\r` in shebangs, not because the code guarantees correct output. **Note: this round's actual live windows-latest CI run already exercised this exact code and passed** (see Blocking #1's real CI evidence above), so this is a real-but-currently-non-manifesting robustness gap, not an active failure — same treatment as prior rounds' latent-risk items. `.just/tests/test_prerelease_archive_workflow.py:117-160`. **Fix (recommended, given the track record on this exact CI mechanism):** write these files with explicit `newline="\n"` so shim content is byte-identical regardless of host line-ending conventions.

5. **(flaky-test-qa, recommendation not a defect)** `_bash()`'s Windows-specific resolver logic has zero direct/mocked unit test — it's only exercised end-to-end via the one Windows CI test. A future edit to this resolver would get no fast local feedback and could silently reintroduce WIN-CI-001 without failing until the next Windows CI round-trip. Recommend a small mocked unit test (monkeypatch `os.name='nt'` and `subprocess.run`, mirroring the existing `PosixShellDiscoveryTests` pattern in `.just/tests/test_transfer_scripts.py`) asserting a System32 candidate is rejected and a real mingw64/usr candidate is accepted. Not required to close this round.

### Recommendation

Route back to `arch-ctm`/fenix for a QA-5 candidate. Priority order: (1) fix `update_manifest_versions()`'s over-inclusive "changed" reporting for Python manifests and correct the regression test that currently locks in the wrong behavior — this is the only Blocking item, it's small and mechanical; (2) delete dead `extract_version_from_url()`; (3) tighten `replace_version_occurrences()`'s trailing boundary guard and add the adjacency test; (4) add `newline="\n"` to the POSIX shim-file writes, given this exact area's track record. Item (5) is optional. **Both of QA-3's real Blocking items (Windows CI, winget URL corruption) are now closed with solid, independently-reproduced evidence — this is real progress, not another overclaim.** The remaining Blocking item is narrow, mechanical, and was caught before merge by three independent readers, which is exactly what this gate is for.

randlee:
## Final Quality Report — QA-5 — PR #1096 `feature/ar1-prerelease-archive`

**Reviewed commit: `2fadf2ff33c7e21d0d4e43fbb08d1cc8ea6ff2dc`** (confirmed real PR head via `gh api repos/randlee/atm-core/pulls/1096`; fix commit `c5fb6cd34` confirmed via `git merge-base --is-ancestor` as a genuine ancestor). Full panel (req-qa, arch-qa, ruthless-boundary-qa, rust-qa-agent, flaky-test-qa), re-verifying closure of all 4 QA-4 findings.

**Verdict: FAIL — not mergeable.** All 4 QA-4 findings are now genuinely fixed, confirmed independently by every applicable reviewer and by me directly. But this round bundled an **unplanned refactor** (extracting the Git-Bash resolver into a shared `resolve_posix_shell()`) that was not on team-lead's stated fix list — and that refactor introduced a real, previously-absent regression in the exact System32-WSL-stub rejection boundary that took 3 rounds (WIN-CI-001) to get right. Two independent reviewers (ruthless-boundary-qa, req-qa) found it separately; I confirmed it myself by diffing the old vs. new resolution logic directly. Per this repo's merge gate (0 Blocking + 0 Important + 0 Minor, no exceptions), this blocks merge on its own.

### Confirmed genuinely fixed (unanimous across all applicable reviewers)

1. **QA5-PYPROJECT-CHANGED (was Blocking)** — `update_manifest_versions()` in `.just/prerelease_tag.py:120-124` now seeds `before` from `[*manifests, *python_projects]` (previously cargo manifests only), captured before `sync_python_version()` runs. The regression test (`test_candidate_bump_updates_actual_lockfile_collision_safely`) now asserts the negative case directly: `assertNotIn(root / "crates" / "atm-graft-python" / "pyproject.toml", changes)`, and that project's `pyproject.toml` was confirmed (by me, req-qa, arch-qa independently) to have `dynamic = ["version"]` with no literal `version =` field — a genuine exercise of the fixed behavior, not a placebo assertion. Confirmed by req-qa, arch-qa, rust-qa-agent, and me.

2. **QA5-DEAD-URL-WRAPPER (was Minor)** — `extract_version_from_url()` deleted from `.just/check_version_sync.py`. Repo-wide grep (req-qa) confirms zero remaining references anywhere. Confirmed by req-qa and me.

3. **QA5-WINGET-URL-BOUNDARY (was Minor)** — `replace_version_occurrences()`'s trailing lookahead changed from `(?![0-9])` to `(?![0-9.])`, now symmetric. New test `test_replace_version_occurrences_rejects_adjacent_version_component` hand-traced independently by req-qa and ruthless-boundary-qa: old_version "1.1.2" inside `"...atm_1.1.2.0_x86_64.zip"` is correctly left unreplaced (the following "." now blocks the match). Ruthless-boundary-qa additionally re-swept both files for the same overmatch class and found no other instance (one unrelated Minor noted below is a different, non-overmatch gap).

4. **QA5-SHIM-LF-NEWLINE (was Minor)** — all 5-6 `write_text()` calls in `_python3_shim`/`_python_command_shim` (`.just/tests/test_prerelease_archive_workflow.py`) now carry `newline="\n"`, confirmed line-by-line by rust-qa-agent and req-qa independently — broader than strictly required (both `.cmd` and POSIX shebang files covered).

5. **Real CI-equivalent local validation** (rust-qa-agent): zero Rust files touched (fresh `git fetch origin develop`, correct merge-base); `cargo fmt --check`/`cargo clippy -D warnings` clean; `cargo test --workspace` 1514 passed/0 failed/2 ignored; `python3 .just/run_pytests.py` 903 passed/11 skipped/0 failed; `just lint` 32/32. My own local run matched: 81/81 targeted tests pass across the 5 touched modules.

6. **QA-4's non-blocking recommendation (direct unit test for the Git-Bash resolver) — satisfied.** Confirmed by flaky-test-qa: the new `test_lint_common.py` tests (`test_resolve_posix_shell_prefers_bash_outside_windows`, `test_resolve_posix_shell_rejects_system32_wsl_stub`, `test_resolve_posix_shell_ignores_ambient_git_bash`) genuinely mock `subprocess.run`/`shutil.which`/`Path.is_file` and drive both OS branches via an explicit `windows=` kwarg — not tautological.

### Blocking

1. **New regression in `_is_system32_wsl_stub` (`.just/lint_common.py:111-114`), introduced by this round's unplanned `resolve_posix_shell()` extraction — independently found by ruthless-boundary-qa and req-qa, verified by me directly.**

   The prior code (removed in commit `c03c97137`, previously in `.just/tests/test_prerelease_archive_workflow.py`) rejected any Git-Bash candidate whose **resolved, symlink-followed** path fell under `System32\` (`Path.resolve()` + `casefold().startswith(system32_prefix + os.sep)` — I confirmed this directly: `git show 4b93114f5:.just/tests/test_prerelease_archive_workflow.py` shows `system32.resolve()` and `candidate.resolve()`). The new code:
   ```python
   def _is_system32_wsl_stub(candidate: str, system32: str) -> bool:
       normalized_candidate = ntpath.normcase(ntpath.abspath(candidate))
       normalized_system32 = ntpath.normcase(ntpath.abspath(system32))
       return normalized_candidate == ntpath.join(normalized_system32, "bash.exe")
   ```
   `ntpath.abspath()` is purely lexical — no filesystem I/O, no symlink/junction resolution (verified by ruthless-boundary-qa empirically) — and the comparison is now **exact-match against the single literal path** `System32\bash.exe`, not a directory-prefix match.

   Two independent, concrete bypass scenarios were found:
   - **ruthless-boundary-qa**: a PATH entry that is a symlink/junction into `System32` (e.g. a `C:\fake` junction to `System32`, or a `Sysnative`-style redirector) — old code followed it via `.resolve()` and correctly rejected it; new code never resolves the junction, so `Path(candidate).is_file()` passes and the WSL stub is silently accepted.
   - **req-qa**: a real, non-symlink candidate at any *subdirectory* of System32 (e.g. `System32\Sub\bash.exe`) — old code's prefix match rejected anything under `System32\`; new code's exact-match only rejects the single literal path `System32\bash.exe`, so this bypasses it even with zero symlinks involved.

   Both reviewers independently confirmed the existing test (`test_resolve_posix_shell_rejects_system32_wsl_stub`) does **not** actually exercise this boundary — its mocked `Path.is_file` only returns `True` for the legitimate Git Bash candidate, so the System32 candidate is filtered out by `is_file()` before `_is_system32_wsl_stub`'s own boundary logic is meaningfully tested.

   This is a real regression in the exact mechanism this PR needed 3 rounds to get right (WIN-CI-001), introduced by scope not on team-lead's declared fix list, and it currently passes CI only because standard `windows-latest` GitHub Actions runners don't happen to exhibit the triggering PATH condition — not because the code is correct. Escalating to Blocking given this PR's established track record in this exact file/area. **Fix:** restore symlink/junction resolution (`Path(candidate).resolve()` or `os.path.realpath`) before comparison, and restore prefix-based rejection (anything resolving under `System32\`, not just the single literal `bash.exe` name) — then add a test that isolates this boundary without the `is_file()` mock short-circuit masking it (e.g. mock `Path.resolve`/`os.path.realpath` to simulate a symlink target, or use a real temp-dir junction on Windows CI).

### Minor

2. **(flaky-test-qa)** The new `GIT_BASH` override assertion in `test_plan_step_exercises_authenticated_release_probe` (`.just/tests/test_prerelease_archive_workflow.py` ~line 250-255) is dead/tautological on non-Windows CI runners: `resolve_posix_shell()` is called with no `windows=` kwarg, so on macOS/Linux it takes the `shutil.which("bash")` fast path regardless of the `mock.patch.dict(os.environ, {"GIT_BASH": ...})` wrapper, making `assertEqual(resolve_posix_shell(), shell)` prove nothing. Not a functional gap (the real Windows logic is genuinely covered by the new mocked `test_lint_common.py` unit tests), but misleading to a future reader. **Fix (non-blocking):** delete the two lines, or make the call meaningful (`resolve_posix_shell(windows=True, git_bash=...)`), or add a comment pointing to where real coverage lives.

### Process note (non-blocking, for the record)

req-qa's full-diff review (`git diff 4b93114f5..2fadf2ff3 --stat`, 12 files / 10 commits) found this push bundled more than the 4 briefed fixes: the `resolve_posix_shell()` extraction (3 commits), a Windows-CI-found test-isolation fix for that same extraction (`e23dac88a`, landed after the fix commit), a subprocess timeout added to the picker test runner, and several doc-only commits recording QA history and disambiguating Phase AS naming. None of it is functionally uninvestigated — every piece has a matching test or is doc-only, and the fix commit `c5fb6cd34` itself is scoped exactly to the 4 declared items. But it's exactly this kind of undisclosed bundled scope that introduced today's Blocking finding — recommend future QA dispatches call out "also includes an incidental refactor of X" explicitly so it gets reviewer attention proportional to its risk, not just its billing.

### Recommendation

Route back to `arch-ctm` for QA-6: fix `_is_system32_wsl_stub` to restore symlink-resolution + prefix-based rejection, add a test that genuinely isolates that boundary, and confirm real Windows CI still passes. The dead `GIT_BASH` assertion cleanup is optional. **All 4 originally-declared QA-5 items are genuinely and solidly fixed** — this FAIL is entirely attributable to new, unplanned scope bundled into the same push, caught before merge by two independent reviewers plus direct verification, which is exactly what this gate exists to do.

randlee:
## Final Quality Report — QA-6 — PR #1096 `feature/ar1-prerelease-archive`

**Reviewed commit: `c5b4dc661c66f07aebaf4ef89ff464f76e8f3da8`** (confirmed real PR head via `gh api repos/randlee/atm-core/pulls/1096`, `mergeable: true`). Full panel (req-qa, arch-qa, ruthless-boundary-qa, rust-qa-agent, flaky-test-qa), verifying closure of the QA-5 `_is_system32_wsl_stub` Blocking regression plus the dead-assertion Minor.

**Verdict: PASS — mergeable.** Both QA-5 findings are genuinely and solidly fixed, confirmed independently by every applicable reviewer and by me directly. No new findings at Blocking, Important, or Minor severity. Merge gate (0B+0I+0m) is met.

### Diff scope (confirmed by 3 independent reviewers + me)

`git diff 2fadf2ff33c7e21d0d4e43fbb08d1cc8ea6ff2dc..c5b4dc661` — exactly 3 files, 37 insertions / 5 deletions: `.just/lint_common.py`, `.just/tests/test_lint_common.py`, `.just/tests/test_prerelease_archive_workflow.py`. No scope creep — this round did not repeat QA-5's undisclosed-bundled-scope problem.

### Blocking item — CONFIRMED-FIXED (`_is_system32_wsl_stub` regression)

New implementation:
```python
def _is_system32_wsl_stub(candidate: str, system32: str) -> bool:
    resolved_candidate = str(Path(candidate).resolve(strict=False)).replace("\\", "/").casefold().rstrip("/")
    resolved_system32 = str(Path(system32).resolve(strict=False)).replace("\\", "/").casefold().rstrip("/")
    return resolved_candidate == resolved_system32 or resolved_candidate.startswith(f"{resolved_system32}/")
```
Restores `Path.resolve(strict=False)` (symlink/junction-following, matching the pre-regression `_bash()` helper's semantics) and a directory-prefix comparison with an explicit separator (closing the nested-subdirectory bypass while avoiding a lookalike-name false positive like `System32Evil`).

Both QA-5 bypass scenarios independently hand-verified closed:
- **Symlink/junction into System32** (ruthless-boundary-qa's finding) — reproduced with a real temp-dir symlink fixture; rejected. `Path.resolve()` empirically follows symlinks on this sandbox; Windows junction-resolution via `Path.resolve()`→`ntpath.realpath` is real CPython behavior (not directly exercisable on macOS, noted as a residual non-blocking platform-coverage gap).
- **Plain nested subdirectory** (`System32\Sub\bash.exe`, req-qa's finding) — rejected via the prefix check.
- **Legitimate outside candidate** (`Git\bin\bash.exe`) — still correctly accepted.
- **Substring/lookalike false-positive check** (`System32Extra\bash.exe`) — correctly NOT rejected; the `rstrip("/")` + `startswith(system32 + "/")` construction is a genuine directory-boundary check, not a naive substring match.

New test `test_system32_wsl_stub_rejection_resolves_prefix_and_symlinks` (`.just/tests/test_lint_common.py`) calls `_is_system32_wsl_stub` **directly** — not routed through `resolve_posix_shell()`, avoiding the `is_file()` mock short-circuit that masked this exact boundary in QA-5's test. Covers literal/nested/symlink/outside cases with real assertions. Executed independently by 3 reviewers (ruthless-boundary-qa, req-qa, flaky-test-qa) and by rust-qa-agent's full suite run — passes for real (not skipped) on this macOS sandbox.

Confirmed by: req-qa, arch-qa, ruthless-boundary-qa, rust-qa-agent, flaky-test-qa, and me directly.

### Minor item — CONFIRMED-FIXED (dead `GIT_BASH` assertion)

The tautological assertion was removed from `test_plan_step_exercises_authenticated_release_probe` and relocated to a new dedicated test, `test_ambient_git_bash_override_does_not_replace_resolved_shell`, gated `@unittest.skipUnless(os.name == "nt", "GIT_BASH override behavior is Windows-only")`. Confirmed genuinely meaningful now (not moved-and-still-broken): `resolve_posix_shell()`'s non-Windows fast path never consults `GIT_BASH`, so gating this assertion to Windows-only is correct rather than cosmetic. Confirmed by req-qa and flaky-test-qa: skips correctly (not falsely "passing" a no-op) on this non-Windows runner, and CI's `just-lint (windows-latest)` job genuinely exercises it there (flaky-test-qa traced `.just/run_lint.py`'s unittest discovery + the CI workflow matrix).

### Architectural fitness (arch-qa)

CONFIRMED-SCOPED, and the fix is architecturally correct, not just functionally correct: it restores `Path.resolve()`-based real-filesystem semantics for the one function whose job is a security-relevant boundary decision (`_is_system32_wsl_stub`), while correctly leaving the syntactic `ntpath`-based candidate-builder helpers (`_windows_shell_candidates`, `_git_bash_candidates`) untouched — those don't need real filesystem resolution, only cross-platform-testable path arithmetic. Single call site confirmed (`resolve_posix_shell`); no other production code duplicates the fixed semantics and needs the same fix.

### Real CI-equivalent local validation (rust-qa-agent)

Zero Rust files touched since fresh `origin/develop` merge-base. `cargo fmt --check` / `cargo clippy --workspace --all-targets -D warnings` clean. `cargo test --workspace`: 1514 passed / 0 failed / 2 ignored. `python3 .just/run_pytests.py`: 904 passed / 12 skipped / 0 failed (baseline was 903/11 — +1 passed, +1 skipped, exactly matching the new test pair). `just lint`: 32/32. Targeted module run (`test_lint_common` + `test_prerelease_archive_workflow`): 35 tests, 34 passed / 1 skipped (the Windows-gated test, correctly skipped) / 0 failed.

### Non-blocking notes (informational only, do not affect verdict)

- **(arch-qa, Low)** The two resolve/normalize lines in `_is_system32_wsl_stub` duplicate the same normalization chain; a tiny local helper would remove it, but this matches the pre-refactor original's style and is cosmetic.
- **(ruthless-boundary-qa)** The `resolved_candidate == resolved_system32` exact-equality branch is real but effectively unreachable through normal candidate-construction paths (which always append `bash.exe`); only reachable via a degenerate `git_bash` override pointing at the System32 directory itself. Makes rejection strictly more conservative, never less — not a bypass risk.
- **(ruthless-boundary-qa)** Windows-native junction/`\\?\` UNC-prefix/8.3-short-path behavior could not be empirically verified on this macOS sandbox; flagged as a residual platform-coverage gap for confirmation on a real Windows CI run, not a demonstrated defect.
- **(flaky-test-qa)** The new symlink test's `self.skipTest(OSError)` fallback is narrow and correct (only triggers on genuine platform/privilege limits, doesn't mask a code regression) — if it does skip on some CI runner, that run loses symlink-branch coverage but cannot produce a false pass.
- **(req-qa)** One unrelated, pre-existing test failure (`tests/test_run_fuzz.py::test_cli_contract_probe_is_a_real_validator_passing_run`, a worktree-path-outside-repo check) reproduces identically at both the QA-5 head and this QA-6 head — confirmed pre-existing/environmental, not caused by this fix, out of this PR's scope.
- **(req-qa)** Sprint doc's test-count references were already stale as of QA-5 and remain so — pre-existing, out of scope for this fix round.

### Recommendation

Merge. All items from QA-1 through QA-6 are closed. No outstanding Blocking, Important, or Minor findings.

PR #1129, opened 2026-09-01
chore: upgrade sc-compose ecosystem to 1.6.1

randlee:
Summary:
- Upgrade the exact standalone CLI, Python binding, and Rust sc-composer/sc-sha pins to 1.6.1.
- Verify downloaded standalone artifacts against official 1.6.1 checksums.
- Make ecosystem release preflight reject a functional but stale sc-compose binary.
- Align bootstrap tests and skill preflight guidance with that contract.

Verification:
- just bootstrap (standalone CLI and Python binding installed at 1.6.1)
- just lint
- just test
- Targeted ecosystem-pin tests, including stale-binary rejection

Note: ecosystem-preflight reaches the new exact-version check. Its only remaining blocker is the pre-existing, unrelated Wyvern currency finding.


rand-lee:
## Windows and Linux CI remediation

### Windows `Just lint`
- **Root cause:** `tools/bootstrap.toml` pinned an incorrect SHA-256 for `sc-compose_1.6.1_x86_64-pc-windows-msvc.zip`. The configured digest differed from the upstream `checksums.txt` digest by one hexadecimal character, so bootstrap rejected the verified release before lint began.
- **Resolution:** commit `51165a916` corrects the Windows release pin and extends the bootstrap unit test to assert the Windows digest explicitly.

### Linux x64 Hermes wheel smoke
- **Root cause:** `sc-compose==1.6.1` was added to the shared `tools/bootstrap-requirements.txt`. The musllinux smoke container installs that generic set before the ATM wheels. `sc-compose` is a native renderer binding with no compatible Alpine/musllinux wheel, so pip attempted a Rust source build and failed before the wheel pair was tested.
- **Resolution:** commit `1eb661f08` removes the renderer-only binding from the generic bootstrap manifest/requirements. The dedicated `setup-renderer` action remains the sole owner of its exact `sc-compose==1.6.1` installation. A regression assertion prevents it being reintroduced into the common bootstrap closure.

### Validation
- `python 3.14 .just/tests/test_bootstrap.py`: pass, 26 tests (1 platform-specific skip).
- Fresh Python 3.14 venv + `pip install --no-deps --requirement tools/bootstrap-requirements.txt`: pass; no native renderer build attempted.
- `just lint`: pass locally; its Python stage reports 918 tests passed, 37 skipped.
- `just test`: not completed locally because this agent execution wrapper terminates foreground commands at two minutes and forbids the background-process workaround. The pushed CI run is the cross-platform completion check.
- Exact Alpine container smoke was not runnable locally because Docker Desktop is not running on this Windows host.

rand-lee:
CI follow-up for commit \1eb661f08\: both repaired gates are green: \Just lint (windows-latest)\ and \Hermes ATM release wheel (musllinux x86_64)\. The corresponding macOS/Ubuntu lint and all Hermes release-wheel/ABI jobs are also green at this check.

PR #1212, opened 2026-09-05
fix(aw-readiness): arch findings M1/M2/M3/M5/O5

randlee:
Phase-aw production-readiness fix round (arch lane): AW-READY-M1, M2, M3, M5, O5.

Branch: fix/aw-readiness-arch -> integrate/phase-aw. Triage records under .triage/phase-aw/findings/.
Dev: arch-ctm. QA: quality-mgr (qa-pr<this>-r1).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_012sVcDUfetCpfiYeQLxYPQK

PR #1211, opened 2026-09-05
fix(phase-aw): diagnostics pagination cap-alignment, admission routing, typed client

randlee:
## Summary

Fixes three phase-aw readiness findings, all scoped to `atm-http-runtime` /
`atm-core` / `atm-storage` (Tokio+Axum stack) — no legacy synchronous daemon
code touched.

### Finding 5 (arch-ctm) / AW-READY-M6 — `/v1/diagnostics` cap drift, no pagination, no deadline
- Introduced a single shared constant, `DIAGNOSTIC_QUERY_MAX_LIMIT` (=1000),
  defined once in `atm-storage` and re-exported through `atm-runtime` and
  `atm-core::observability_counters`. The route, the sqlite storage clamp,
  and the CLI (`atm log`) all now enforce the same cap — previously the
  route advertised up to 5000 while storage silently clamped to 1000.
- Added opaque cursor-based (keyset) pagination over `(ts_unix_ms, id)`,
  stable across same-millisecond ties. `DiagnosticCursor` encodes/decodes a
  base64 URL-safe-no-pad token; malformed cursors return 400.
- Added an additive `truncated: bool` + `next_cursor: Option<String>` to the
  response (`#[serde(default)]`, so existing clients are unaffected).
- Wrapped the storage query in `tokio::time::timeout(request_timeout, ...)`
  so a slow/blocked query now fails closed with 503 instead of hanging
  indefinitely.

### C13 / AW-READY-O7 — auxiliary routes bypass the admission layer
`/v1/health` and `/v1/diagnostics` previously bypassed the load-shed +
concurrency-limit layer applied to canonical write routes. Built a second,
independently-sized admission stack (same `HandleErrorLayer` +
`LoadShedLayer` + `ConcurrencyLimitLayer` composition, reusing the existing
`overload_response` handler) and applied it via `route_layer` to the merged
health+diagnostics router before merging into the canonical router. The
canonical write-route admission stack itself was left untouched to avoid any
hot-path risk.

### C11 / AW-READY-O7 — untyped `loopback_tcp_get_json` escape hatch
Narrowed `loopback_tcp_get_json` to `pub(crate)` and added a typed
`diagnostics_query(...)` client method as the one public, testable entry
point for the diagnostics endpoint. Migrated the CLI's only consumer
(`atm log`'s `timeline_records_from_endpoint`) onto the typed client and
removed its duplicated `percent_encode` helper. Also fixed a latent bug this
[...]

PR #1215, opened 2026-09-05
fix: harden AW readiness findings

randlee:
Closes assigned AW readiness findings on the Tokio/Axum phase branch.

Implemented in isolated commits:
- AW-READY-P1: preserve same-share file sources and verify copy bytes (`f2962ecc1`)
- AW-READY-M4: prune oldest diagnostic rows under count cap (`2603bbe56`)
- AW-READY-M7: rank diagnostic levels numerically (`7f678a558`)
- AW-READY-M8: escape component LIKE filters and document bootstrap fragility (`7d1790d56`)
- AW-READY-O7 assigned subset: C7/C8/C9/C12/C14/C15 (`75cf105e9`, `55d75a732`, `2ad8c8eaa`, `298326647`, `f0d109b88`, `adeffcb4b`)

Validation: cargo fmt --all -- --check; direct boundary lint; workspace clippy with -D warnings; requested package test suites.

C10/C11/C13 remain with their owning branches; no legacy daemon files were touched.

### DIFF
--- a/.just/lint-config.toml
+++ b/.just/lint-config.toml
@@ -158,7 +158,7 @@ allowed_dependencies = ["async-trait", "atm-core", "atm-daemon-client", "atm-htt
 
 [[boundaries.manifest_dependency_allowlists]]
 owner_manifest_path = "crates/atm-graft-python/Cargo.toml"
-allowed_dependencies = ["atm-core", "atm-graft", "atm-observability", "pyo3", "serde_json", "tempfile", "tokio"]
+allowed_dependencies = ["atm-core", "atm-graft", "pyo3", "serde_json", "tempfile", "tokio"]
 
 [[boundaries.manifest_dependency_allowlists]]
 owner_manifest_path = "crates/atm-query-python/Cargo.toml"
--- a/.just/tests/test_lint_boundaries.py
+++ b/.just/tests/test_lint_boundaries.py
@@ -20,6 +20,7 @@ from lint_boundaries import IO_FORBIDDEN_SOURCE_PATTERNS
 from lint_boundaries import parse_boundary_records
 from lint_boundaries import parse_simple_yaml_document
 from lint_boundaries import run
+from run_lint import build_tasks
 
 
 ROOT_MANIFEST = """\
@@ -199,6 +200,17 @@ notes = []
 
 
 class LintBoundariesTests(unittest.TestCase):
+    def test_declared_peer_dial_lint_rule_resolves_to_registered_runner(self) -> None:
+        records, parse_violations = parse_boundary_records(REPO_ROOT)
+        self.assertEqual(parse_violations, [])
+        runtime = next(
+            record
+            for record in records
+            if record.source_path == Path("boundaries/atm-http-runtime/http-runtime.toml")
+        )
+        self.assertIn("peer-dial-seam", runtime.lint_rules)
+        self.assertIn("peer-dial-seam", build_tasks(REPO_ROOT))
+
     def test_graft_shared_client_has_no_direct_interprocess_dependency(self) -> None:
         manifest = tomllib.loads(
             (REPO_ROOT / "crates/atm-graft/Cargo.toml").read_text(encoding="utf-8")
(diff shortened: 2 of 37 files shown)


randlee:
## qa-pr1215-r5 verdict: FAIL -- confirmed real regression, not a CI flake

Re-pinned to 9e6383f0b17b459da176456bafba2a5fa6a1af04 per the post-rebase stack. `git diff fee774b96..9e6383f0b` is identical to the #1211/#1212 delta: only `.triage/phase-aw/findings/AW-READY-W1.ttl` (+16 lines, Rand's phase-ay deferral ruling). The `test_cli_parity.py` regression that failed CI at `fee774b96` is byte-for-byte unchanged at the new head, so it will fail again on this SHA.

**Root cause (confirmed via code review, not just the CI log):** `AW-READY-N5` (`9a036c50f`, "test: isolate CLI parity runtime scope") isolates the CLI-subprocess side of `test_cli_parity.py` (via a `HOME`/`ATM_HOME`/`ATM_TEST_RUNTIME_HOME` env dict passed only to `subprocess.run(...)` calls) but never applies that same environment to the test's own process before constructing the in-process native binding:

- `crates/atm-graft-python/tests/test_cli_parity.py` `_GeneratedParityFixture.start()` builds an isolated `environment` dict and returns it as `fixture["environment"]`.
- `CliParityTests.setUpClass` sets `cls.environment = cls.fixture["environment"]`, used only inside `cls._cli(...)` (subprocess env for CLI comparisons).
- Immediately after, `cls.native_tools = native_tools.AtmNativeTools(identity=..., team=..., chat_id=...)` is constructed with no home override and no `os.environ` mutation -- it resolves against the real, unmodified process environment.

So the CLI subprocess and daemon are isolated under a disposable temp root, but the native binding used for the "parity" comparison looks at the real host's `~/.atm` (or a bare CI-runner home with nothing in it) for team `aw5-cli-parity`, which was never created there. Its `atm_list` call returns an error envelope with no `"result"` key -- exactly the observed `KeyError: 'result'`. This is independent of debug/release build mode (I initially suspected the `#[cfg(debug_assertions)]` gate on `home.rs`'s `ATM_TEST_RUNTIME_HOME` override, since the native side loads a release wheel built by `hermes-atm-release-wheels`/maturin-action --release -- that gate is real and would also break this, but the deeper bug is that `os.environ` is never updated for the test's own process at all, so even a debug-built native binding would fail the same way).

**Suggested fix (for whoever picks this up, likely arch-ctm):** in `setUpClass`, apply the fixture's isolated environment to the test process itself (e.g. `os.environ.update(cls.environment)` with restoration in `tearDownClass`, or add an explicit home-override parameter to `AtmNativeTools`'s constructor) before constructing `native_tools`/`second_native_tools`, so both sides of the parity comparison resolve the same runtime scope. The downstream `atm doctor --json` failure in `run_thorough_shared_host.py` in the same CI job is very likely a cascade of this (leftover/mismatched daemon-socket state from the failed run), not an independent second regression -- not independently confirmed, flag for whoever fixes this to re-check once the primary bug is fixed.

**#1211/#1212 unaffected**: their post-rebase deltas are the identical single-commit triage-ttl addition, no touch to N5/test infra -- no new review needed, PASS stands from round-3 at their new heads.

**Verdict: FAIL for #1215 at 9e6383f0b** (blocking -- breaks CI `test_cli_parity.py` on all three platforms). PASS re-pin for #1211 @ 09698970e and #1212 @ 0287eff30 (unchanged scope).

https://claude.ai/code/session_01MRTztdpYFQmSMYbNwHdNtL

randlee:
qa-final-repin (qa-pr1215-r7): independently re-verified at ee943e2d1516829a6972711b0fc6fce0cce3f84d. Diff from 123e5a3af77bffcdf86ee053bb4b935453e6d720 (the SHA I already fully re-verified this round: rebuilt native atm_graft extension from source via maturin develop, ran the previously-failing test directly — `ATM_CLI_PARITY_CI=1 python3 -m unittest crates.atm-graft-python.tests.test_cli_parity -v` → 1 test, OK — plus `cargo fmt --check --all` clean) is exactly one file: .triage/phase-aw/findings/AW-N5-ENV-001.ttl (+42, the deferred-triage record documenting this very fix). No code/test changes since my from-scratch verification. Ancestry confirmed: fix/aw-readiness-arch@c90c36dd3 is an ancestor. AW-N5-ENV-001 fix (os.environ isolation in setUpClass/tearDownClass before constructing AtmNativeTools) is retained and independently confirmed working. **Verdict: PASS at ee943e2d1516829a6972711b0fc6fce0cce3f84d.** Per team-lead's freeze ruling, this is the final review round for this stack; any further minor/important findings go to fix/aw-readiness-followup, not this branch.

PR #1222, opened 2026-09-05
fix(runtime): keep remote receiver hooks peer-owned

randlee:
Fixes the cross-host sender warning by skipping local receiver-hook emission after a host-qualified local write is handed to its peer.

The receiver-side peer ingress remains the sole hook owner. Adds a loopback peer regression covering a sender without a local recipient roster entry.

Validation: `just ci`, `cargo test --workspace`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -375,7 +375,8 @@ jobs:
         shell: bash
         env:
           ATM_CLI_PARITY_CI: "1"
-        run: "$ATM_PARITY_PYTHON" -m unittest discover -s crates/atm-graft-python/tests -p "test_cli_parity.py"
+        run: |
+          "$ATM_PARITY_PYTHON" -m unittest discover -s crates/atm-graft-python/tests -p "test_cli_parity.py"
 
       - name: Run ATM queue lifecycle hook tests
         if: ${{ needs.ci-scope.outputs.runtime == 'true' }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -350,12 +350,32 @@ jobs:
           python -m pip install --disable-pip-version-check --requirement tools/bootstrap-requirements.txt
           python -m pip install --no-deps "$ATM_GRAFT_WHEEL" crates/hermes-atm
 
+      - name: Install isolated debug CLI/native parity packages
+        if: ${{ needs.ci-scope.outputs.runtime == 'true' }}
+        shell: bash
+        run: |
+          set -euo pipefail
+          venv_root="$GITHUB_WORKSPACE/.bootstrap-venv"
+          if [[ "$RUNNER_OS" == "Windows" ]]; then
+            venv_python="$venv_root/Scripts/python.exe"
+            venv_bin="$venv_root/Scripts"
+          else
+            venv_python="$venv_root/bin/python"
+            venv_bin="$venv_root/bin"
+          fi
+          export VIRTUAL_ENV="$venv_root"
+          export PATH="$venv_bin:$PATH"
+          "$venv_python" -m pip install --no-deps crates/hermes-atm
+          maturin develop --manifest-path crates/atm-graft-python/Cargo.toml
+          "$venv_python" -c 'import atm_graft._atm_graft as extension; assert hasattr(extension, "_debug_runtime_scope")'
+          echo "ATM_PARITY_PYTHON=$venv_python" >> "$GITHUB_ENV"
+
       - name: Run CLI/native parity test with a generated fixture
         if: ${{ needs.ci-scope.outputs.runtime == 'true' }}
         shell: bash
         env:
           ATM_CLI_PARITY_CI: "1"
-        run: python -m unittest discover -s crates/atm-graft-python/tests -p "test_cli_parity.py"
+        run: "$ATM_PARITY_PYTHON" -m unittest discover -s crates/atm-graft-python/tests -p "test_cli_parity.py"
 
       - name: Run ATM queue lifecycle hook tests
         if: ${{ needs.ci-scope.outputs.runtime == 'true' }}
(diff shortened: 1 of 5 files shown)


randlee:
qa-pr1222-r1: independently reviewed at c8319d4b30514e832840061758879b821f2972fc (fix/aw-readiness-followup, based on the frozen fix/aw-readiness-easy @ ee943e2d1 — merge-base confirmed). One commit, two files:

- crates/atm-http-runtime/src/storage_and_nudge_router.rs: hoists the host-qualified-recipient check into a single `is_remote_recipient` computed once, keeps the peer-dispatch condition (`committed.newly_persisted && is_remote_recipient`) unchanged, and gates the sender-local received hook on `committed.newly_persisted && !is_remote_recipient` — so a host-qualified send that's already been handed to its peer no longer also fires (and fails to resolve against) the sender's own local roster. Matches the described bug exactly: a misleading local-hook warning despite peer delivery and the peer's own hook succeeding.
- New/extended regression test drops the sender-side local roster entry entirely (`fixture(false, ...)`) and asserts: the send still succeeds with no warnings, the sender's local hook never fires (empty `emitted_ids`), and the remote peer's hook fires exactly once. This is a real behavioral assertion, not just a smoke check.
- crates/atm-architecture/tests/boundary_enforcement.rs: one-line source-string match update (`if committed.newly_persisted` vs the old `if committed.newly_persisted {`) to track the refactor — no boundary-rule relaxation.

No crates/atm-daemon/ touch. Independently ran `cargo test -p atm-http-runtime -p atm-architecture` (209 + doctests + the boundary test, all pass) and `cargo fmt --check --all` (clean) from scratch, not just trusting arch-ctm's self-report. **Verdict: PASS at c8319d4b30514e832840061758879b821f2972fc.**

randlee:
## quality-mgr verdict: FAIL — new commit 3147244f2e44d6ec1a9514777dff37f8e10b7764

My prior PASS at `c8319d4b3` no longer applies; the head moved without a re-pin request. This commit ("fix(graft): align release wheel parity runtime") is a genuine regression, not a fix.

**What it does**: removes `ATM_TEST_RUNTIME_HOME` from `_GeneratedParityFixture._environment()` in `crates/atm-graft-python/tests/test_cli_parity.py`, with the stated rationale that release-wheel artifacts ignore this debug-only override and the debug daemon would otherwise honor it, splitting client/daemon across two runtime scopes.

**Why that's wrong**: `current_host_runtime_scope()` (`crates/atm-core/src/home.rs`) is `test_runtime_home()?.unwrap_or(os_account_home()?)`. On Unix, `os_account_home()` calls `libc::getpwuid(libc::geteuid())` — it resolves the **real OS account home** and explicitly ignores `HOME`/`ATM_HOME` by design (see the function's own comment: "this intentionally does not use HOME, ... those are process-scoped inputs and cannot define a host-wide singleton boundary"). Removing `ATM_TEST_RUNTIME_HOME` doesn't align the debug CLI/daemon with the release wheel's scope — it removes the *only* mechanism that redirected the debug CLI/daemon pair out of the real account home. The `HOME`/`ATM_HOME` overrides also set in this fixture's environment dict do nothing for `current_host_runtime_scope()`.

**Confirmed impact (reproduced twice, then verified against the real store)**:
- `ATM_CLI_PARITY_CI=1 python3 -m unittest crates.atm-graft-python.tests.test_cli_parity -v` fails on a second run (fresh temp dir each time) with `RuntimeError: ... 'teams', 'add-member', 'aw5-cli-parity', 'aw5-parity-sender' ...: member 'aw5-parity-sender' already exists in team 'aw5-cli-parity'`.
- Directly confirmed in the real production store: `/Users/randlee/.atm/db/mail.db` now contains rows for team `aw5-cli-parity`, members `aw5-parity-sender`/`aw5-parity-receiver`, and a persisted "parity reply" message dated `2026-09-05T21:15:30Z` from this test run. `atm teams` against the real account home lists `aw5-cli-parity (2)` as a live team.

This is the same defect class as AV-HOTFIX-002 (Rand's ruling: "tests must never run against the live atm database... every test touching the trust store (or any store) must run under an isolated home"), reintroduced by this commit rather than fixed by it.

**Required fix**: restore `ATM_TEST_RUNTIME_HOME` in the fixture. If release-wheel vs. debug-daemon scope alignment is a real problem, it needs a different mechanism (e.g. the release-wheel side also honoring an isolated scope override, or a documented decision from Rand to accept a split scope for the mandatory parity lane specifically) — not deleting the isolation boundary.

Blocking. Also flagging for cleanup: the real `~/.atm` store on this host now has fixture team/member/message rows from `aw5-cli-parity` that should be purged once the fix lands.

randlee:
## quality-mgr verdict: PASS — qa-pr1222-r2 at 9d9bcd0b1e0a2fe3c7f83ead1415e9861bcc9c0

Independently confirmed arch-ctm's safety-restoration commit reverts the regression I caught at `3147244f2` cleanly.

1. **Fixture matches pre-regression state exactly**: `git diff c8319d4b3 HEAD -- crates/atm-graft-python/tests/test_cli_parity.py` is empty — the file is byte-identical to the SHA I originally PASSed. No release-wheel workaround, no `environment.pop("ATM_TEST_RUNTIME_HOME", ...)`, no leftover `GeneratedParityFixtureTests` test, and no unused `mock` import remain.
2. **Isolation verified effective, not just present**: `ATM_TEST_RUNTIME_HOME` is back in `_environment()`. Ran `ATM_CLI_PARITY_CI=1 python3 -m unittest crates.atm-graft-python.tests.test_cli_parity -v` twice in a row (the regression only surfaced on a second run previously) — both pass. Checked `strings /Users/randlee/.atm/db/mail.db | grep -c aw5-cli-parity` before and after both runs: **44 both times** — the pre-existing contamination count from the earlier regression is unchanged; no new writes landed in the real store. (Per team-lead: those pre-existing 44 rows are left in place pending Rand's cleanup decision — not touched here.)
3. **Standard gates clean**: `cargo fmt --check --all` clean. `just lint` — 36/36 checks pass including pytests.

Noted team-lead's caveat: local `just ci` doesn't exercise the release-wheel parity lane, so it isn't evidence for that lane. I didn't rely on it — the isolation check above is a direct repro of the exact failure mode from the FAIL report (`ATM_CLI_PARITY_CI=1 python3 -m unittest ...`, twice, with before/after DB grep), independent of `just ci`.

PASS.

PR #1223, opened 2026-09-05
fix(tooling): support Windows daemon-switch lifecycle

rand-lee:
### Root cause
The existing Windows lifecycle path used `sc.exe`. ATM runs per-user and derives state from the user's `ATM_HOME`; an SCM service normally runs as LocalSystem, so it cannot safely manage the user-owned daemon or its state. This left Windows `daemon-switch` unsupported in practice.

### Resolution
- Manage the Windows daemon through a per-user interactive Scheduled Task instead of SCM.
- Add `windows-provision` to create the logon task with the configured daemon selector.
- Verify that the task action points exactly at the configured selector before starting it; reject direct worktree binaries and absent/mismatched tasks.
- Add Windows status, start, and stop handling through `schtasks.exe`.
- Fail `temporary-launch` closed on Windows because the Scheduled Task backend intentionally owns lifecycle.
- Document the Windows provisioning and selector requirements.

### Validation
- `py -3.12 .claude/skills/daemon-switch/tests/test_daemon_switch.py WindowsScheduledTaskTests` (5 passed)
- Live verification on FastPC4: provisioned `atm-daemon`, switched the paired selectors to installed ATM 1.5.0, and `daemon-switch.py status --doctor` reported a matched 1.5.0 CLI/daemon pair with a healthy daemon.

The all-platform test module contains pre-existing POSIX-only cases (`AF_UNIX`, `os.getuid`, POSIX permissions); this PR adds only Windows-specific lifecycle coverage and does not change Unix service paths.

### DIFF
--- a/.claude/skills/daemon-switch/SKILL.md
+++ b/.claude/skills/daemon-switch/SKILL.md
@@ -90,22 +90,36 @@ binary identifier and certificate common name. A changed or mismatched leaf
 fails closed. Apple-issued `Apple Development` identities continue to use the
 team-identifier verification path.
 
-## Windows selector provisioning
+## Windows selector and task provisioning
 
-The script never replaces ordinary `.exe` files. Provision two user-writable
-selector symlinks named `atm.exe` and `atm-daemon.exe` in one directory placed
-before the installed release directory on `PATH`, then pass them explicitly:
+The script never replaces ordinary `.exe` files. Windows uses one
+**interactive-user Scheduled Task**, not an SCM service: SCM services commonly
+run as `LocalSystem` and would select a different `ATM_HOME` and durable store.
+Provision two user-writable selector symlinks named `atm.exe` and
+`atm-daemon.exe` in one directory placed before the installed release directory
+on `PATH`, then register the task to invoke the daemon selector:
 
 ```powershell
+python .claude/skills/daemon-switch/scripts/daemon-switch.py windows-provision `
+  --cli-link C:\\atm-active\\atm.exe --daemon-link C:\\atm-active\\atm-daemon.exe `
+  --service atm-daemon --yes
+
 python .claude/skills/daemon-switch/scripts/daemon-switch.py switch `
   --cli-link C:\\atm-active\\atm.exe --daemon-link C:\\atm-active\\atm-daemon.exe `
   --cli target\\release\\atm.exe --daemon target\\release\\atm-daemon.exe --yes `
   --service atm-daemon
 ```
 
-Creating those symlinks may require Developer Mode or an elevated shell. If
-they are unavailable, the script fails closed; do not replace installed
-executables or introduce a second daemon.
+`windows-provision` creates an on-logon, interactive-user task whose only
+executable action is `C:\\atm-active\\atm-daemon.exe`. `switch`, `restart`,
+and `quiesce` refuse a task whose action names a worktree or any other direct
+binary. Corporate policy may require running only `windows-provision` from an
+elevated terminal; do not work around that policy by registering an SCM service
+or launching a second daemon.
+
+The typed `temporary-launch` overlay is not supported by this task backend.
+It fails closed instead of modifying a task action outside its scoped
+pair-switch contract.
 
 ## Default scratch root (ADR-055)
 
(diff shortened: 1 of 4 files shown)


randlee:
## quality-mgr verdict: FAIL — qa-pr1223-r1 at e624b0471

**Reviewers**: req-qa (FAIL, 2 Blocking) + arch-qa (PASS, 3 Important). I independently re-verified the contested Blocking claim before relying on it (see below) — it's real, not a stale-doc false positive.

### Blocking (from req-qa, independently confirmed)
1. **Unapproved removal of a currently-Accepted requirement**: `docs/requirements.md:311-314` (REQ-P-DAEMON-SWITCH-001) and ADR-053 (Status: Accepted) both explicitly mandate that the Windows temporary-launch adapter "capture the named SCM service's exact `BINARY_PATH_NAME`... and restore the exact captured scalar through SCM." I read both documents directly — the requirement is current, not stale, and ADR-053 is the authoritative design baseline for this exact control plane. This PR's `temporary_launch_adapter()` (`.claude/skills/daemon-switch/scripts/daemon-switch.py:117-135`) now unconditionally raises `SwitchError` for Windows instead of returning `WindowsScmAdapter`, and SKILL.md documents this as permanent. No corresponding requirements.md/ADR-053 amendment accompanies the change.
2. Same root cause, second citation: ADR-053's "Required evidence" section states Windows completeness cannot rest on "fake boundaries alone" — directly contradicted by removing the Windows adapter path entirely.

This isn't necessarily wrong as an engineering decision (the code's own comment — "do not substitute an SCM service because it would run under a different account" — is a real, defensible reason a per-user scheduled-task backend can't safely reuse the old SCM-based temporary-launch overlay). But it's a capability removal against a currently-Accepted ADR/requirement, same class as the standing "unapproved breaking wire change is blocking" rule (ADR-061) applied here to a control-plane requirement instead of the wire schema. This needs either (a) Windows temporary-launch support restored some other way, or (b) an explicit ADR-053/REQ-P-DAEMON-SWITCH-001 amendment approved before merge — not a silent drop.

### Important (from arch-qa, independently spot-checked, not blocking this gate)
- **ARCH-001**: the Linux/generic branch (`daemon-switch.py:546-548`) has the *same* bug class fenix's fix (`e624b0471`) just closed for Windows — `if result.returncode == 0 or (allow_absent and action == "stop"): return` treats any non-zero `systemctl` exit as "already absent," including permission-denied. I confirmed this directly in the current code. It's pre-existing (not a new regression from this PR, so it satisfies Rand's "no behavior change for Linux" ask), but it's directly adjacent to the bug just fixed and worth a fast follow-up.
- **ARCH-002**: `WindowsScmAdapter` (`temporary_launch_windows.py`) is now dead in the dispatch path (only reachable via direct unit-test instantiation) — needs an explicit disposition (delete, or document as ADR-053-pending reference-only) rather than being left ambiguous. This is entangled with the Blocking item above — resolving that will resolve this too.
- **ARCH-003**: the new access-denied regression test (`.just/tests/test_daemon_switch.py:315`) wasn't ported to the larger, canonical `.claude/skills/daemon-switch/tests/test_daemon_switch.py` suite, and separately that suite isn't even wired into `.just/run_pytests.py`'s discovery — so it doesn't run in `just test`/CI at all (req-qa's ATM-QA-003/004, same finding from both reviewers independently).

**Everything else** (windows_task_status/run_service access-denied handling — the actual scope of fenix's fix, `provision_windows_task` `--yes`/`/RL LIMITED` gating and no-credential-leak, SKILL.md accuracy, workspace-wide sweep confirming no other `sc.exe` reference remains in live code) is confirmed correct by both reviewers plus my own `just lint` (36/36) run at this SHA.

**Gate**: FAIL — blocking item 1 needs Rand's ruling before merge (restore capability vs. approve ADR-053 amendment). Routing to fenix per the assignment's instruction (cwin is dispatched only via Rand).

randlee:
Re-pin only (stack rebase, linear now): head is now `b74f3aa5d`. Scoped diff against prior pinned head `e624b0471` is empty for this layer's own content. Verdict unchanged: **PASS**.

randlee (approved):

PR #1227, opened 2026-09-05
fix(ci): AW-PARITY-ISOLATION-1 — debug atm-graft for CLI parity lane

randlee:
Stacked on #1222 (fix/aw-readiness-followup). Layer 5 of the AW readiness stack.

Release builds compile out `ATM_TEST_RUNTIME_HOME`, so with runtime isolation restored in 9d9bcd0b1 the release atm-graft wheel and the debug daemon split sockets and the parity lane fails. CI keeps the release-wheel install/smoke and additionally installs a debug atm-graft into the bootstrap venv for behavioural parity. Generated-fixture setup now calls the debug-only runtime-scope helper and hard-fails before native-session creation unless it resolves under the fixture temp home.

Dev validation (arch-ctm, 9ca526035): parity PASS with `ATM_CLI_PARITY_CI=1`; `just lint` all checks pass; no `~/.atm` touched.

https://claude.ai/code/session_012sVcDUfetCpfiYeQLxYPQK

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -375,7 +375,8 @@ jobs:
         shell: bash
         env:
           ATM_CLI_PARITY_CI: "1"
-        run: "$ATM_PARITY_PYTHON" -m unittest discover -s crates/atm-graft-python/tests -p "test_cli_parity.py"
+        run: |
+          "$ATM_PARITY_PYTHON" -m unittest discover -s crates/atm-graft-python/tests -p "test_cli_parity.py"
 
       - name: Run ATM queue lifecycle hook tests
         if: ${{ needs.ci-scope.outputs.runtime == 'true' }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -350,12 +350,32 @@ jobs:
           python -m pip install --disable-pip-version-check --requirement tools/bootstrap-requirements.txt
           python -m pip install --no-deps "$ATM_GRAFT_WHEEL" crates/hermes-atm
 
+      - name: Install isolated debug CLI/native parity packages
+        if: ${{ needs.ci-scope.outputs.runtime == 'true' }}
+        shell: bash
+        run: |
+          set -euo pipefail
+          venv_root="$GITHUB_WORKSPACE/.bootstrap-venv"
+          if [[ "$RUNNER_OS" == "Windows" ]]; then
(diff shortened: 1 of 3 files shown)


randlee:
## quality-mgr qa-pr1227-r1 verdict — PASS

Pinned to `366dd7141` (re-pin from `9ca526035`: adds a ci.yml YAML quoting fix — nested quoted run scalar → block scalar — no behavior change to the parity-isolation logic itself).

**Reviewers:** req-qa PASS (4/4 deliverables, 0 findings), arch-qa PASS (0 blocking, 0 important).

Independently verified alongside the reviewers:
1. **No release-user semantic change**: `_debug_runtime_scope()` in `crates/atm-graft-python/src/lib.rs` is `#[cfg(debug_assertions)]`-gated on both the function definition and its pymodule registration; a release `maturin build --release` compiles it out entirely, and it's underscore-prefixed so it's excluded from `atm_graft`'s public `import *` surface even in debug builds.
2. **No silent fallback to release wheel or real `~/.atm`**: `ci.yml` installs the debug `atm-graft` build via `maturin develop` into `.bootstrap-venv` (separate from the release wheel's system-Python install, no PATH/env leakage across step boundaries), exports `ATM_PARITY_PYTHON`, and the parity test step invokes that interpreter directly. `_assert_debug_native_scope()` runs in `setUpClass` strictly before native-session creation and raises `RuntimeError` (not skip/warn) on a missing env var, a missing debug symbol (release wheel), or a scope mismatch.
3. **Hard-fail is real**: confirmed via direct read — three raise paths, no soft-skip branch.
4. **No legacy sync daemon coupling introduced**: the diff only touches `ci.yml` + `lib.rs` + `test_cli_parity.py`'s guard; the parity fixture's pre-existing use of `target/debug/atm-daemon` predates this diff and falls under the legacy-daemon exemption (frozen, Phase-AM deletion target) — not remediated here, correctly out of scope.
5. Verdict based on code review, not gated on CI (merge gate only).

Two non-blocking observations from arch-qa, awareness only, no action required: `.bootstrap-venv` now double-duties as both the pinned-tooling cache and the test-fixture Python install target (minor concern-mixing, optional future cleanup); legacy-daemon pairing in the fixture is pre-existing and exempt.

**Gate: PASS, 0 Blocking, 0 Important.** This is the top of the AW readiness stack (#1211→#1212→#1215→#1222→#1227) — PASS here plus green CI at this head is the stack merge signal.

PR #1283, opened 2026-09-07
fix(graft): nudge carries the rendered template only (REQ-CORE-GRAFT-001)

### LINKED ISSUE #1281: graft nudge carries the full message body (REQ-CORE-GRAFT-001 violation, introduced by #987)
Since PR #987 (commit 94d59d257, 2026-08-22) the graft post-send payload carries the admitted message body in addition to the rendered nudge:

- crates/atm-core/src/send/hook.rs passes message text into build_built_in_dispatch and copies it into GraftNudgeTarget.message_body
- crates/atm-core/src/boundary/mod.rs GraftNudgeTarget.message_body (doc comment: "so an embedded host never needs to perform a follow-up mailbox read")
- crates/atm-core/src/graft.rs GraftPostSendRequest.message_body (wire contract to every graft receiver)
- crates/atm-graft/src/nudge_sink.rs injects rendered_nudge + "\n\n" + message_body into the host session

Effects observed on the live hermes team (mail store, 2026-09-07 round trip): hermes agents act on the pushed body and never call atm_read, so mail_message_states.read stays 0 for every graft-delivered message. Unread counts inflate permanently and the ADR-043 restart advisory is built on them.

Requirements violated: REQ-CORE-GRAFT-001 (graft is a bounded host-wake transport, not a second mailbox; read goes through the daemon API), the internal-nudge envelope contract (event, sink target, template kind, template body only), and ADR-043 item 2 (daemon mailbox is the sole authority for unread state). Rand's ruling: the emitter is backend-blind; every sink gets the same rendered template and either steer or queue; content crosses only via read.

GH-985's "body truncated" report was a mis-diagnosis of the intended envelope-only behaviour; the receiver-rebind half of #987 stays.

Fix: remove message_body from GraftNudgeTarget and GraftPostSendRequest, drop the message-text argument from build_built_in_dispatch, revert nudge_sink to inject only the resolved template, add a regression test that the graft request and injected nudge contain no message text.

randlee:
## Summary

Fixes #1281

Both graft and bare-CLI QueuePull post-send paths now carry only the canonical rendered nudge template. Neither receiver path transports an additional admitted message-body payload, preserving REQ-CORE-GRAFT-001, REQ-CORE-COMPAT-003, ADR-043, and the one-emitter rule.

## Changes

- Removed message_body from GraftNudgeTarget and GraftPostSendRequest.
- Removed the message-text parameter from the built-in dispatch helper.
- Bare-CLI QueuePull now renders the built-in template for its dispatch kind and stores that rendered XML in QueuePullTarget.body and QueuedNudgeMessage.body.
- Added graft emitter and bare-CLI dispatch regressions asserting template-only payloads.
- Updated graft sink, queue selector, FIFO, and HTTP queue-get fixtures/assertions.

## QueueGetNext body semantics

QueueGetNextResponse.body for bare-CLI harness consumers now contains the rendered nudge XML, not admitted message text. Consumers must use the nudge message ID and run atm read to retrieve the mailbox message. This is the intentional semantics change required by the same defect.

## Wire compatibility and rollout

The 1.5.3 graft receiver expects message_body, so this fix ships as a paired 1.5.4 daemon and wheel release. Gateways must restart after the wheel is installed. No serde shim or default empty body is used.

## Validation

All required validation passes: cargo fmt, workspace clippy with warnings denied, focused package tests, full workspace tests, function-length, boundary lint, and line-count checks.

### DIFF
--- a/.winget/randlee.agent-team-mail.yaml
+++ b/.winget/randlee.agent-team-mail.yaml
@@ -1,5 +1,5 @@
 PackageIdentifier: randlee.agent-team-mail
-PackageVersion: 1.5.3
+PackageVersion: 1.5.4
 PackageLocale: en-US
 Publisher: Randy Lee
 PublisherUrl: https://github.com/randlee
@@ -21,7 +21,7 @@ Tags:
 Installers:
   - Architecture: x64
     InstallerType: zip
-    InstallerUrl: https://github.com/randlee/atm-core/releases/download/v1.5.3/atm_1.5.3_x86_64-pc-windows-msvc.zip
+    InstallerUrl: https://github.com/randlee/atm-core/releases/download/v1.5.4/atm_1.5.4_x86_64-pc-windows-msvc.zip
     InstallerSha256: <SHA256_HASH_TO_BE_FILLED_AFTER_RELEASE>
 ManifestType: installer
-ManifestVersion: 1.5.3
+ManifestVersion: 1.5.4
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,7 +4,7 @@ version = 4
 
 [[package]]
 name = "agent-team-mail"
-version = "1.5.3"
+version = "1.5.4"
 dependencies = [
  "agent-team-mail-core",
  "anyhow",
@@ -34,7 +34,7 @@ dependencies = [
 
 [[package]]
 name = "agent-team-mail-core"
-version = "1.5.3"
+version = "1.5.4"
 dependencies = [
  "async-trait",
  "atm-error",
@@ -182,7 +182,7 @@ dependencies = [
 
 [[package]]
 name = "atm-architecture"
-version = "1.5.3"
+version = "1.5.4"
 dependencies = [
  "cargo_metadata",
  "quote",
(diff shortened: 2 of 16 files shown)


randlee:
## Review (fenix, release lane) — PASS

Scope matches Rand's ruling and GH-1281: the post-persist emitter is backend-blind; every sink gets the rendered template only.

Verified on head 872918841 against develop 9654b75f1:
- `build_built_in_dispatch` no longer takes message text; both callers in `nudge_dispatch.rs` and `write/pipeline.rs` pass the event only.
- `GraftNudgeTarget` and `GraftPostSendRequest` lose `message_body`; new emitter test pins the serialized graft request keys to exactly `event`, `kind`, `rendered_nudge`.
- `atm-graft` sink injects the rendered template only (test renamed and asserts no message text).
- Bare-CLI queue-pull branch now renders the template for its kind and carries it in `QueuePullTarget.body` / `QueuedNudgeMessage.body`; FIFO tests assert `<atm` bodies and no message text.
- No `message_body` remains in boundary, graft, atm-graft, or the Python bindings.
- Receiver-rebind half of #987 untouched. No legacy daemon changes.

Wire-compat note accepted as written: 1.5.3 receivers require the removed field, so the fix ships as the 1.5.4 daemon + wheel pair with gateway restart. No shim.

Merge gate: CI green on this head. The 1.5.4 version bump is being added to this same head to avoid a second CI cycle.

Issue #893, opened 2026-08-16
Native atm_read mutates read state, violates 'without mutating it' contract


randlee:
## Summary

The native `atm_read` tool (registered by the `hermes-atm-native-tools` plugin) marks messages as read and returns `mutation_applied=true`, contradicting its documented contract of "Read one ATM mailbox message **without mutating it**."

## Root cause

In `crates/atm-graft-python/src/query.rs`, `build_tool_read_query` constructs a `ReadQuery` with `seen_state_update=false`, which only suppresses the seen-watermark update — NOT the read-state mutation.

In `crates/atm-core/src/read/mod.rs`, the read path calls `read_mail_with_runtime_impl` with `DisplayMutationMode::MutatingRead`, which always transitions unread messages to read via `apply_display_mutations_to_store`. The non-mutating peek path (`peek_mail_with_runtime_impl` → `DisplayMutationMode::NonMutatingPeek`) is never reached from the native `read_tool`.

The call chain: `read_tool` → `read_message` → `RequestEnvelope::Receive` → `read_mail_with_runtime_impl(MutatingRead)`.

## Affected behavior

- **Expected:** native `atm_read(selection='all', message_id=...)` returns `mutation_applied=false` and does NOT set `read=1` on the message.
- **Actual:** returns `mutation_applied=true` and the message is marked read in the mailbox state.

## False-positive from prior tests

If the target message is already read (read=1), `displayed_messages_require_mutation` returns `false` even on the `MutatingRead` path, so `mutation_applied` appears `false`. This can give a false impression that the read is non-mutating when it's actually just idempotent on already-read messages. A fresh unread self-send (as in the SELFACK release proof) catches this.

## Fix

Route the native `read_tool` to the non-mutating peek path (`peek_mail_with_runtime_impl` → `NonMutatingPeek`), or add a non-mutating request envelope to the daemon API.

## Evidence

- RC test: ALPHA_PRIME-RC-SELFACK-20260816-01, step 3
- Self msg `01M0426MNADARZS984SJVBAZGG` was read=0 before native read, read=1 after
- DB query confirmed the state transition

randlee:
Verified fixed on develop.

**Evidence:**
- Commit cd59d6b77 (2026-08-15): "fix(graft): make native reads non-mutating"
- Commit references Finding HERMES-RC-002-READ-MUTATION
[...]

PR #1287, opened 2026-09-07
fix(graft-python): native atm_read marks the message read; atm-daemon --version (1.5.5)

### LINKED ISSUE #1286: fix(graft-python): native atm_read must mark the message read (supersedes #893); atm-daemon --version
## Summary

The Hermes native `atm_read` tool never sets `read = 1`. Commit cd59d6b77 (PR #896, issue #893, finding HERMES-RC-002) routed `read_tool` to the non-mutating peek path so the tool would match its own description text ("Read one ATM mailbox message without mutating it"). Rand's ruling 2026-09-07: **the read flag is set to read when a read occurs.** The description was wrong, not the read path. `atm_read` is the native equivalent of `atm read` and must follow §7.5 / §7.11 of docs/requirements.md: any selected message is written back with `read = true`; ack state is untouched.

Live evidence: on the hermes team every graft-delivered message stays `read = 0` even when the agent replies and acks (see #1281 for the nudge-body half of that observation, fixed in #1283 / 1.5.4). With #1283 in place the agents will call `atm_read`, and it will still leave `read = 0` because of this issue.

## Not a write-queue concern

`R-READ-CONC-1` / `R-STATE-HANDOFF-1` (§7.13) already keep reads off the writer lane: the read is served by the concurrent reader lane and the read/seen transition is a bounded, non-blocking handoff that the read path never awaits (ADR-059). `R-STATE-RACE-1` makes the observed state race-tolerant. Restoring the mutating read for the native tool adds one handoff per read; it does not serialize reads behind writes. CLI `atm read` already works this way.

## Required change

1. `crates/atm-graft-python/src/lib.rs` `read_tool`: use the mutating read operation (`ReadOperation::Read` / `client.read_message`), not `peek_outcome`. `mutation_applied` reports the real handoff result.
2. `crates/atm-graft-python/src/query.rs` / `tool_types.rs`: drop the `Peek` routing and `from_peek_outcome` that exist only for the native tool; keep `atm peek` semantics for the CLI untouched.
3. `crates/hermes-atm/src/hermes_atm/native_tools.py`: description of `atm_read` becomes "Read one ATM mailbox message and mark it read; ack state is unchanged. Use the ATM CLI for administrative or advanced operations." Add a native `atm_peek`? No: out of scope; the CLI has `atm peek`.
4. Flip the tests from #896 that assert `mutation_applied == false` / message stays unread for the native read; add one test proving `(Unread, PendingAck) -> (Read, PendingAck)` through `read_tool` (§7.11).
5. `crates/atm-daemon/src/main.rs`: `atm-daemon --version` must print `atm-daemon <CARGO_PKG_VERSION>` and exit 0 before bootstrap. Today it ignores the flag and tries to start (fails on the singleton lock when a daemon is live, would start a second daemon otherwise). This blocks `daemon-switch.py switch --worktree`, which reads both binary versions, so no locally built pair can be switched in.
6. Patch bump to 1.5.5 on develop and tag `prerelease/v1.5.5` at the merge commit.

## Requirements

- docs/requirements.md §7.5 `atm read` required behavior, §7.11 Mutation Rules, §7.13 read-lane concurrency and read-state handoff; ADR-059; ADR-043 item 2 (daemon mailbox sole authority for read state).
- Supersedes the direction taken in #893.

randlee:
## Summary

Fixes #1286.

- Native `atm_read` now performs the canonical mutating read by default, marking the selected message read through the existing bounded non-blocking state handoff; acknowledgement state is unchanged.
- Native `atm_read` accepts `peek=true` for the existing non-mutating inspection path.
- This supersedes #893 by Rand’s 2026-09-07 ruling: the native read description must match the required read semantics.
- `atm-daemon --version` and `-V` now print `atm-daemon 1.5.5` before daemon bootstrap, even when another daemon is live.
- Workspace and package metadata are bumped to 1.5.5.

## Requirements and boundaries

The read behavior follows docs/requirements.md sections 7.5, 7.11, and 7.13 and ADR-059: the reader lane serves the response, the read transition is offered to the existing bounded handoff without awaiting the writer lane, and acknowledgement state is untouched. CLI `atm read` and `atm peek` paths are unchanged.

## Tests

- Focused Rust tests: `cargo test -p atm-graft-python -p atm-graft -p atm-daemon -p agent-team-mail-core`
- Graft Python harness: `python3 scripts/test_atm_graft_python.py`
- Hermes bridge wheel and tests: `python3 .just/run_hermes_graft_bridge_tests.py` (19 tests)
- Model schema tests: `python3 .just/tests/test_atm_graft_python_models.py`
- Full workspace: `cargo test --workspace`
- Static and release gates: fmt, clippy, function-length, boundaries, line counts, metadata, and `atm-daemon --version`

All listed gates pass.

### DIFF
--- a/.just/tests/test_atm_graft_python_models.py
+++ b/.just/tests/test_atm_graft_python_models.py
@@ -78,6 +78,7 @@ class AtmGraftPythonModelTests(unittest.TestCase):
                 "contains",
                 "since",
                 "from_agent",
+                "peek",
             },
             models.AtmListRequest: {
                 "selection",
--- a/.winget/randlee.agent-team-mail.yaml
+++ b/.winget/randlee.agent-team-mail.yaml
@@ -1,5 +1,5 @@
 PackageIdentifier: randlee.agent-team-mail
-PackageVersion: 1.5.4
+PackageVersion: 1.5.5
 PackageLocale: en-US
 Publisher: Randy Lee
 PublisherUrl: https://github.com/randlee
@@ -21,7 +21,7 @@ Tags:
 Installers:
   - Architecture: x64
     InstallerType: zip
-    InstallerUrl: https://github.com/randlee/atm-core/releases/download/v1.5.4/atm_1.5.4_x86_64-pc-windows-msvc.zip
+    InstallerUrl: https://github.com/randlee/atm-core/releases/download/v1.5.5/atm_1.5.5_x86_64-pc-windows-msvc.zip
     InstallerSha256: <SHA256_HASH_TO_BE_FILLED_AFTER_RELEASE>
 ManifestType: installer
-ManifestVersion: 1.5.4
+ManifestVersion: 1.5.5
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,7 +4,7 @@ version = 4
 
 [[package]]
 name = "agent-team-mail"
-version = "1.5.4"
+version = "1.5.5"
 dependencies = [
  "agent-team-mail-core",
  "anyhow",
@@ -34,7 +34,7 @@ dependencies = [
 
 [[package]]
 name = "agent-team-mail-core"
-version = "1.5.4"
+version = "1.5.5"
 dependencies = [
  "async-trait",
  "atm-error",
@@ -182,7 +182,7 @@ dependencies = [
 
 [[package]]
 name = "atm-architecture"
-version = "1.5.4"
+version = "1.5.5"
 dependencies = [
  "cargo_metadata",
  "quote",
@@ -193,7 +193,7 @@ dependencies = [
 
 [[package]]
 name = "atm-daemon"
-version = "1.5.4"
+version = "1.5.5"
 dependencies = [
  "agent-team-mail-core",
  "atm-daemon-bootstrap",
@@ -202,7 +202,7 @@ dependencies = [
 
 [[package]]
 name = "atm-daemon-bootstrap"
-version = "1.5.4"
+version = "1.5.5"
 dependencies = [
  "agent-team-mail-core",
  "atm-herdr",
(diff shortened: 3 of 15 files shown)


randlee:
## fenix review — 675892f0f

Verified against issue #1286 and Rand's rulings (2026-09-07):

- **Default `atm_read` is the canonical mutating read.** `build_tool_read_query` returns `ReadOperation::Read` (ReadQuery, seen-state update on, activity observation attached) unless `peek=true`, which keeps the pre-existing non-mutating `PeekQuery` route. Read is still served on the reader lane; the Unread→Read transition goes through the existing writer handoff and is never awaited (R-STATE-HANDOFF-1). Nothing new lands on the writer lane.
- **Ack axis untouched.** Test `native_read_marks_messages_read_without_changing_ack_state` covers both `(Unread, PendingAck) → (Read, PendingAck)` and `(Unread, NoAckRequired) → (Read, NoAckRequired)`; `native_read_does_not_reapply_an_already_read_transition` covers one-directionality; `native_read_peek_option_leaves_the_message_unread` covers the exception path.
- **Tool contract.** `AtmReadRequest.peek: bool = False` added to the Pydantic ingress model; `atm_read` description now states that the message is marked read and that `peek=true` inspects without marking. The "without mutating it" wording from #893 is gone.
- **`atm-daemon --version` / `-V`** prints `atm-daemon 1.5.5` and exits 0 before bootstrap; unit test pins the string. This unblocks `daemon-switch.py switch --worktree` for local builds.
- **Version bump 1.5.5** across the same files as the 1.5.4 bump (Cargo.toml/lock, pyproject files, winget manifest).

Supersedes the direction of #893. No findings. Merge gate: CI green on this head (checks in progress at 07:14Z).

PR #1299, opened 2026-09-07
fix: restore native loopback reliability for 1.5.7

### LINKED ISSUE #1297: Native loopback client: stale pooled keep-alive connection turns every request after ~3 s idle into ATM_DAEMON_MAY_HAVE_EXECUTED
## Summary
Every native (hermes-atm / atm-graft) request over the loopback HTTP client fails with `ATM_DAEMON_MAY_HAVE_EXECUTED` when the previous request on the same session was more than ~3 s ago. The failure is instantaneous (0.00 s), the daemon never sees the request (nothing in `atm.log.jsonl`), and an immediate retry succeeds. Reproduced on 1.5.5 and 1.5.6 (daemon f20441cc7). It is the cause of the SMOKE155/SMOKE156 native send/ack/list failures reported by loki, paternostrier and skillrx (#1289 items 1–2; skillrx saw it on `atm_list`, so it is not write-specific).

## Reproduction (runtime-v4 venv, 1.5.6 wheels, daemon pid 85105)
`PyGraftSession(PyAgentAddress("fenix","atm-dev","probe"))`, one `send_tool` to an invalid recipient per step (the validation error proves the daemon handled the request):

| idle before request | result | elapsed |
| --- | --- | --- |
| warm (0 s) | ATM_MESSAGE_VALIDATION_FAILED (daemon reached) | normal |
| 2 s | ATM_MESSAGE_VALIDATION_FAILED (daemon reached) | normal |
| 4 s | **ATM_DAEMON_MAY_HAVE_EXECUTED** | 0.00 s |
| immediate follow-up | ATM_MESSAGE_VALIDATION_FAILED (daemon reached) | normal |
| 6 s / 8 s / 10 s | **ATM_DAEMON_MAY_HAVE_EXECUTED**, follow-up OK each time | 0.00 s |

## Mechanism
- Server: `crates/atm-http-runtime/src/http1_server.rs` `serve_connection` builds hyper with `.header_read_timeout(header_read_timeout)` and `.keep_alive(true)`, where `header_read_timeout` is `timeouts.request` (the 3 s request budget). On a keep-alive connection the header timer also bounds the wait for the *next* request, so the daemon closes every idle connection after ~3 s. The code comment assumes "callers send `Connection: close` when their finite batch is complete".
- Client: `crates/atm-http-runtime/src/client.rs` `build_loopback_reqwest_client` uses reqwest's default connection pool (90 s idle) and never sends `Connection: close`. The next request is written on the half-closed pooled socket.
- Classification: `client_tail.rs` `classify_reqwest_execute_failure` maps every non-`is_connect()` error to `HttpRuntimeClientFailure::RequestWrite`, which `into_atm_error` turns into `daemon_may_have_executed(...)` and `is_safe_to_reconnect` rejects. Nothing was ever written to the daemon, so the uncertainty is false.

## Fix direction (arch-ctm to choose, one screen, loopback path only)
Either stop pooling on the loopback client (`pool_max_idle_per_host(0)` or a `pool_idle_timeout` below the server's header timer, or `Connection: close` as the server comment assumes), or give the server a keep-alive idle timeout that is not the request budget, or both. A stale-pool write must never be reported as MAY_HAVE_EXECUTED. Regression test against a real `atm-daemon` (not a scripted fixture): idle > server header timeout, then a write, must succeed. CI currently excludes atm-daemon tests (#1292), so the test has to run where the daemon runs.

## Rollout
Patch bump to 1.5.7 on develop, tag at the tested commit, rollout and SMOKE157 per the standing procedure. Related: #1289, #1291, #1292.

### LINKED ISSUE #1298: Native atm_read by message_id returns count=0 for ids the same session's list just returned
## Summary
Native `atm_read` by `message_id` never finds the message. `PyGraftSession.read_tool(selection, message_id=<id>)` returns `count=0, match=None, mutation_applied=False` for `all`, `unread` and `actionable`, positional or keyword, with or without `peek`, for ids that `list_tool` on the *same session* just returned and that `atm read --message-id <id> --json` (CLI, same identity) resolves with `count=1, mutation_applied=true`. Reproduced on 1.5.6 (daemon f20441cc7, wheels 1.5.6) as fenix@atm-dev; paternostrier reported the same `count=0` on SMOKE156.

`read_tool` signature on 1.5.6: `(selection="actionable", message_id=None, task=None, contains=None, since=None, from_agent=None, peek=False)`.

## Effect
Every hermes agent's native read-by-id is a no-op, so the read mark from Rand's 2026-09-07 ruling ("read marks read") is never applied by the native tool; the read=1 marks observed on SMOKE155/156 came from agents falling back to the CLI. `mutation_applied=false` in the smoke reports is this defect, not the handoff race in #1291.

## Expectations
- Native read by id resolves the same message the CLI resolves, marks it read (queued handoff, R-STATE-HANDOFF-1), and reports `count=1`, `mutation_applied=true` unless `peek=true`.
- Regression test against a real daemon: list → take a returned id → read by id → count 1, then DB/list shows read.
- Confirm whether the native path applies the session `chat_id` as a filter on the read-by-id selection; the CLI has no chat id and finds the row.

Related: #1289 item 3, #1291. Fix ships with the same 1.5.7 patch as the loopback pool defect.

randlee:
## Summary

- Fixes #1297: both same-host reqwest connectors apply the same bounded two-second idle-pool timeout, below the daemon three-second HTTP/1 header timer. This retains short keep-alive reuse while preventing reuse of a connection the daemon may have closed. Direct-peer pooling and its post-write uncertainty contract are unchanged.
- Fixes #1298: exact-ID reads and peeks resolve the caller-owned mailbox row without a chat-session filter. Agent/team mailbox ownership remains enforced.
- R3 made an intentional behavior change, not a pure refactor: native exact-ID peek now removes PeekQuerys bare-agent chat restriction so an owned message resolves regardless of session chat ID. R4 aligns CLI peek with that ruling. R5 proves the CLI behavior with a different destination session and states the shared query contract once on ReadQuery.
- The HTTP runtime does not select transports. Canonical ATM_LOCAL_TRANSPORT selection remains in atm-daemon-client.
- All release metadata remains at 1.5.7.

## Exact-ID ruling and disposition

An exact message ID in the callers own mailbox is resolved by ownership only, never by chat scope, identically for read and peek, CLI and native.

| Finding | Disposition |
| --- | --- |
| QA-B1 / QA-I2 | Closed: one two-second bounded idle-pool policy covers loopback TCP and UDS. |
| QA-I1 | Closed: CLI exact-ID read bypasses session-chat scope. |
| QA-3-B1 / QA-3-I2 / QA4-I1 | Closed: native and CLI exact-ID peek remove chat scope intentionally; the CLI regression uses a message addressed to a different session and fails with AssertionError: 0 != 1 if the R4 hunk is reverted. |
| QA-3-I1 / QA4-I2 | Closed: the corrected with_caller_chat_id contract appears once on ReadQuery; it states that None clears the participant filter for an exact ID, leaving ownership as the recipient scope. |
| QA-001 | Superseded by the architecture ruling: runtime parser removed; no runtime selection branch remains. |
| RSH-001 / RSH-003 | Closed/accepted: canonical resolver records unset versus empty override state; the construction-time metadata read is isolated and documented. |

## Real-daemon regressions

The generated disposable daemon fixture leaves ATM_LOCAL_TRANSPORT unset. It proves default-UDS idle-write recovery and that native and CLI exact-ID read and peek each resolve an owned message regardless of session chat scope.

## Validation

- cargo fmt --check — PASS
- cargo clippy --workspace --all-targets -- -D warnings — PASS
- cargo test --workspace --quiet — PASS, including atm-daemon
- python3 .just/run_lint.py function-length, lines, boundaries, and version — PASS
- rebuilt debug binding plus ATM_CLI_PARITY_CI=1 parity suite — PASS, 6 tests

Before each push, the PR diff was reviewed against f20441cc7. Each change is exercised by the corresponding production path or regression test.

### DIFF
--- a/.winget/randlee.agent-team-mail.yaml
+++ b/.winget/randlee.agent-team-mail.yaml
@@ -1,5 +1,5 @@
 PackageIdentifier: randlee.agent-team-mail
-PackageVersion: 1.5.6
+PackageVersion: 1.5.7
 PackageLocale: en-US
 Publisher: Randy Lee
 PublisherUrl: https://github.com/randlee
@@ -21,7 +21,7 @@ Tags:
 Installers:
   - Architecture: x64
     InstallerType: zip
-    InstallerUrl: https://github.com/randlee/atm-core/releases/download/v1.5.6/atm_1.5.6_x86_64-pc-windows-msvc.zip
+    InstallerUrl: https://github.com/randlee/atm-core/releases/download/v1.5.7/atm_1.5.7_x86_64-pc-windows-msvc.zip
     InstallerSha256: <SHA256_HASH_TO_BE_FILLED_AFTER_RELEASE>
 ManifestType: installer
-ManifestVersion: 1.5.6
+ManifestVersion: 1.5.7
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,7 +4,7 @@ version = 4
 
 [[package]]
 name = "agent-team-mail"
-version = "1.5.6"
+version = "1.5.7"
 dependencies = [
  "agent-team-mail-core",
  "anyhow",
@@ -34,7 +34,7 @@ dependencies = [
 
 [[package]]
 name = "agent-team-mail-core"
-version = "1.5.6"
+version = "1.5.7"
 dependencies = [
  "async-trait",
  "atm-error",
@@ -182,7 +182,7 @@ dependencies = [
 
 [[package]]
 name = "atm-architecture"
-version = "1.5.6"
+version = "1.5.7"
 dependencies = [
  "cargo_metadata",
  "quote",
@@ -193,7 +193,7 @@ dependencies = [
 
 [[package]]
 name = "atm-daemon"
-version = "1.5.6"
+version = "1.5.7"
 dependencies = [
  "agent-team-mail-core",
  "atm-daemon-bootstrap",
@@ -202,7 +202,7 @@ dependencies = [
 
 [[package]]
 name = "atm-daemon-bootstrap"
-version = "1.5.6"
+version = "1.5.7"
 dependencies = [
  "agent-team-mail-core",
  "atm-herdr",
@@ -226,7 +226,7 @@ dependencies = [
 
 [[package]]
 name = "atm-daemon-client"
-version = "1.5.6"
+version = "1.5.7"
 dependencies = [
  "agent-team-mail-core",
  "atm-http-runtime",
@@ -241,7 +241,7 @@ dependencies = [
 
 [[package]]
 name = "atm-error"
-version = "1.5.6"
+version = "1.5.7"
 dependencies = [
  "serde",
  "serde_json",
@@ -249,7 +249,7 @@ dependencies = [
 
 [[package]]
 name = "atm-graft"
-version = "1.5.6"
+version = "1.5.7"
 dependencies = [
  "agent-team-mail-core",
  "async-trait",
@@ -264,7 +264,7 @@ dependencies = [
 
 [[package]]
 name = "atm-graft-python"
-version = "1.5.6"
+version = "1.5.7"
 dependencies = [
  "agent-team-mail-core",
  "atm-graft",
@@ -276,7 +276,7 @@ dependencies = [
 
 [[package]]
 name = "atm-herdr"
-version = "1.5.6"
+version = "1.5.7"
 dependencies = [
  "agent-team-mail-core",
  "serde_json",
@@ -285,7 +285,7 @@ dependencies = [
 
 [[package]]
 name = "atm-http-runtime"
-version = "1.5.6"
+version = "1.5.7"
 dependencies = [
  "agent-team-mail-core",
  "async-trait",
@@ -312,7 +312,7 @@ dependencies = [
 
 [[package]]
 name = "atm-observability"
-version = "1.5.6"
+version = "1.5.7"
 dependencies = [
  "agent-team-mail-core",
  "sc-observability",
@@ -326,7 +326,7 @@ dependencies = [
 
 [[package]]
 name = "atm-peer-tls-interop"
-version = "1.5.6"
+version = "1.5.7"
 dependencies = [
  "agent-team-mail-core",
  "atm-storage",
@@ -337,7 +337,7 @@ dependencies = [
 
 [[package]]
 name = "atm-query-python"
-version = "1.5.6"
+version = "1.5.7"
 dependencies = [
  "atm-storage",
  "atm-storage-rusqlite",
@@ -347,7 +347,7 @@ dependencies = [
 
 [[package]]
 name = "atm-runtime"
-version = "1.5.6"
+version = "1.5.7"
 dependencies = [
  "agent-team-mail-core",
  "async-trait",
@@ -360,7 +360,7 @@ dependencies = [
 
 [[package]]
 name = "atm-runtime-test-support"
-version = "1.5.6"
+version = "1.5.7"
 dependencies = [
  "agent-team-mail-core",
  "async-trait",
@@ -371,7 +371,7 @@ dependencies = [
 
 [[package]]
 name = "atm-storage"
-version = "1.5.6"
+version = "1.5.7"
 dependencies = [
  "async-trait",
  "atm-error",
@@ -390,7 +390,7 @@ dependencies = [
 
 [[package]]
 name = "atm-storage-rusqlite"
-version = "1.5.6"
+version = "1.5.7"
 dependencies = [
  "async-trait",
  "atm-storage",
@@ -405,14 +405,14 @@ dependencies = [
 
 [[package]]
 name = "atm-storage-sqlserver-proof"
-version = "1.5.6"
+version = "1.5.7"
 dependencies = [
  "atm-storage",
 ]
 
 [[package]]
 name = "atm-template-sc-compose"
-version = "1.5.6"
+version = "1.5.7"
 dependencies = [
  "agent-team-mail-core",
  "atm-storage",
@@ -1476,7 +1476,7 @@ dependencies = [
 
 [[package]]
 name = "peer-tls"
-version = "1.5.6"
+version = "1.5.7"
 dependencies = [
  "atm-storage",
  "rcgen",
@@ -1956,7 +1956,7 @@ dependencies = [
 
 [[package]]
 name = "sc-lint-boundary"
-version = "1.5.6"
+version = "1.5.7"
 dependencies = [
  "anyhow",
  "cargo_metadata",
@@ -1964,7 +1964,7 @@ dependencies = [
  "proc-macro2",
  "quote",
  "sc-lint-attributes",
- "sc-lint-directives 1.5.6",
+ "sc-lint-directives 1.5.7",
  "serde",
  "serde_json",
  "syn 2.0.119",
@@ -1983,7 +1983,7 @@ dependencies = [
 
 [[package]]
 name = "sc-lint-directives"
-version = "1.5.6"
+version = "1.5.7"
 dependencies = [
  "quote",
  "syn 2.0.119",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,7 @@ members = [
 resolver = "2"
 
 [workspace.package]
-version = "1.5.6"
+version = "1.5.7"
 edition = "2024"
 rust-version = "1.94.1"
 authors = ["atm-core contributors"]
@@ -63,17 +63,17 @@ getrandom = "0.3"
 syn = { version = "2", features = ["full", "parsing", "visit"] }
 semver = { version = "1", features = ["serde"] }
 async-trait = "0.1"
-atm-core = { package = "agent-team-mail-core", path = "crates/atm-core", version = "1.5.6" }
-atm-observability = { path = "crates/atm-observability", version = "1.5.6" }
-atm-daemon-bootstrap = { path = "crates/atm-daemon-bootstrap", version = "1.5.6" }
-atm-daemon-client = { path = "crates/atm-daemon-client", version = "1.5.6" }
-atm-error = { path = "crates/atm-error", version = "1.5.6" }
-atm-graft = { path = "crates/atm-graft", version = "1.5.6" }
-atm-herdr = { path = "crates/atm-herdr", version = "1.5.6" }
-atm-http-runtime = { path = "crates/atm-http-runtime", version = "1.5.6" }
-atm-runtime = { path = "crates/atm-runtime", version = "1.5.6" }
+atm-core = { package = "agent-team-mail-core", path = "crates/atm-core", version = "1.5.7" }
+atm-observability = { path = "crates/atm-observability", version = "1.5.7" }
+atm-daemon-bootstrap = { path = "crates/atm-daemon-bootstrap", version = "1.5.7" }
+atm-daemon-client = { path = "crates/atm-daemon-client", version = "1.5.7" }
+atm-error = { path = "crates/atm-error", version = "1.5.7" }
+atm-graft = { path = "crates/atm-graft", version = "1.5.7" }
+atm-herdr = { path = "crates/atm-herdr", version = "1.5.7" }
+atm-http-runtime = { path = "crates/atm-http-runtime", version = "1.5.7" }
+atm-runtime = { path = "crates/atm-runtime", version = "1.5.7" }
 atm-runtime-test-support = { path = "crates/atm-runtime-test-support" }
-atm-storage = { path = "crates/atm-storage", version = "1.5.6" }
-atm-storage-rusqlite = { path = "crates/atm-storage-rusqlite", version = "1.5.6" }
-atm-template-sc-compose = { path = "crates/atm-template-sc-compose", version = "1.5.6" }
-peer-tls = { path = "crates/peer-tls", version = "1.5.6" }
+atm-storage = { path = "crates/atm-storage", version = "1.5.7" }
+atm-storage-rusqlite = { path = "crates/atm-storage-rusqlite", version = "1.5.7" }
+atm-template-sc-compose = { path = "crates/atm-template-sc-compose", version = "1.5.7" }
+peer-tls = { path = "crates/peer-tls", version = "1.5.7" }
--- a/crates/atm-core/src/read/request.rs
+++ b/crates/atm-core/src/read/request.rs
@@ -263,9 +263,6 @@ impl PeekQuery {
     /// A present chat ID scopes a read-only mailbox query to that session.
     /// `None` removes chat scoping, leaving mailbox ownership as the only
     /// recipient scope.
-    /// A present chat ID scopes a read-only mailbox query to that session.
-    /// `None` removes chat scoping, leaving mailbox ownership as the only
-    /// recipient scope.
     #[must_use]
     pub fn with_caller_chat_id(mut self, caller_chat_id: Option<ChatId>) -> Self {
         self.mailbox.participant_filter =
@@ -391,6 +388,9 @@ impl ReadQuery {
     pub fn caller_chat_id(&self) -> Option<&ChatId> {
         self.caller_chat_id.as_ref()
     }
+    /// A present chat ID scopes a read-only mailbox query to that session.
+    /// For an exact ID in the caller's own mailbox, `None` clears the
+    /// participant filter so ownership is the only recipient scope.
     #[must_use]
     pub fn with_caller_chat_id(mut self, caller_chat_id: Option<ChatId>) -> Self {
         self.mailbox.participant_filter =
--- a/crates/atm-core/src/read/request.rs
+++ b/crates/atm-core/src/read/request.rs
@@ -260,9 +260,12 @@ impl PeekQuery {
         self.caller_chat_id.as_ref()
     }
 
-    /// Scope a read-only mailbox query to the caller's identity. A present
-    /// chat ID selects that exact session; `None` selects only the bare-agent
-    /// identity.
+    /// A present chat ID scopes a read-only mailbox query to that session.
+    /// `None` removes chat scoping, leaving mailbox ownership as the only
+    /// recipient scope.
+    /// A present chat ID scopes a read-only mailbox query to that session.
+    /// `None` removes chat scoping, leaving mailbox ownership as the only
+    /// recipient scope.
     #[must_use]
     pub fn with_caller_chat_id(mut self, caller_chat_id: Option<ChatId>) -> Self {
         self.mailbox.participant_filter =
--- a/crates/atm-daemon-client/src/local_transport.rs
+++ b/crates/atm-daemon-client/src/local_transport.rs
@@ -54,8 +54,14 @@ impl LocalDaemonTransport {
 pub fn local_daemon_transport() -> Result<LocalDaemonTransport, AtmError> {
     let value = std::env::var("ATM_LOCAL_TRANSPORT").ok();
     let transport = LocalDaemonTransport::resolve(value.as_deref())?;
+    let override_state = match value.as_deref() {
+        None => "unset",
+        Some("") => "explicitly_empty",
+        Some(_) => "explicit",
+    };
     tracing::debug!(
         transport = transport.label(),
+        local_transport_override = override_state,
         "selected daemon local transport"
     );
     Ok(transport)
--- a/crates/atm-graft-python/src/lib.rs
+++ b/crates/atm-graft-python/src/lib.rs
@@ -1966,6 +1966,49 @@ mod tests {
     }
 
     #[test]
+    fn native_exact_message_id_peek_ignores_session_chat_scope() {
+        let caller = PyAgentAddress::new(
+            TEST_RECIPIENT.to_string(),
+            TEST_TEAM.to_string(),
+            Some("recipient-session".to_owned()),
+        )
+        .expect("caller");
+        let session = PyGraftSession {
+            caller: caller.to_typed().expect("typed caller"),
+            client: Mutex::new(None),
+            receiver: Mutex::new(None),
+            fallback_logger: Arc::new(observability::GraftFallbackLogger::new(
+                atm_core::observability::graft_fallback_log_path(&leaked_fallback_dir()),
+            )),
+            reconnect_replacement: Mutex::new(None),
+            reconnect_attempts: AtomicUsize::new(0),
+            reconnect_fallback_attempts: AtomicUsize::new(0),
+        };
+
+        let operation = session
+            .build_tool_read_query(
+                "all",
+                Some("01KRFK5QTF2R6NRS3Q0F8Z9K0S"),
+                None,
+                None,
+                None,
+                None,
+                true,
+            )
+            .expect("native peek query");
+        let crate::query::ReadOperation::Peek(query) = operation else {
+            panic!("peek option must build a non-mutating query")
+        };
+        let query_json = serde_json::to_value(&query).expect("serialized query");
+
+        assert_eq!(
+            query_json["mailbox"]["message_id_filter"],
+            serde_json::json!("01KRFK5QTF2R6NRS3Q0F8Z9K0S")
+        );
+        assert!(query_json["mailbox"]["participant_filter"].is_null());
+    }
+
+    #[test]
     fn native_read_marks_messages_read_without_changing_ack_state() {
         Python::initialize();
         let outcomes = Arc::new(Mutex::new(vec![
--- a/crates/atm-graft-python/src/query.rs
+++ b/crates/atm-graft-python/src/query.rs
@@ -63,6 +63,12 @@ impl PyGraftSession {
                 )))
             })?;
         let selection = Self::read_selection(selection)?;
+        // `list_tool` and the CLI resolve an exact message ID within the
+        // caller's mailbox scope, irrespective of the host session's chat
+        // identifier. Preserve that parity for native exact-ID reads: the
+        // session-chat filter is appropriate for bucket scans, but would hide
+        // a bare-agent-addressed message that list just returned.
+        let exact_message_id = message_id.is_some();
         if peek {
             PeekQuery::new(
                 home_dir,
@@ -81,7 +87,12 @@ impl PyGraftSession {
             )
             .map_err(atm_error)
             .map(|query| {
-                ReadOperation::Peek(query.with_caller_chat_id(self.caller.chat_id().cloned()))
+                let query = if exact_message_id {
+                    query
+                } else {
+                    query.with_caller_chat_id(self.caller.chat_id().cloned())
+                };
+                ReadOperation::Peek(query)
             })
         } else {
             ReadQuery::new(
@@ -102,14 +113,14 @@ impl PyGraftSession {
             )
             .map_err(atm_error)
             .map(|query| {
-                ReadOperation::Read(
+                let query = if exact_message_id {
                     query
-                        .with_caller_chat_id(self.caller.chat_id().cloned())
-                        .with_activity_observation(activity_observation_for_resolved_caller(
-                            self.caller.agent(),
-                            &team,
-                        )),
-                )
+                } else {
+                    query.with_caller_chat_id(self.caller.chat_id().cloned())
+                };
+                ReadOperation::Read(query.with_activity_observation(
+                    activity_observation_for_resolved_caller(self.caller.agent(), &team),
+                ))
             })
         }
     }
--- a/crates/atm-graft-python/src/query.rs
+++ b/crates/atm-graft-python/src/query.rs
@@ -63,9 +63,8 @@ impl PyGraftSession {
                 )))
             })?;
         let selection = Self::read_selection(selection)?;
-        // `list_tool` and the CLI resolve an exact message ID within the
-        // caller's mailbox scope, irrespective of the host session's chat
-        // identifier. Preserve that parity for native exact-ID reads: the
+        // Exact-ID reads resolve within the caller's mailbox scope,
+        // irrespective of the host session's chat identifier. The
         // session-chat filter is appropriate for bucket scans, but would hide
         // a bare-agent-addressed message that list just returned.
         let exact_message_id = message_id.is_some();
--- a/crates/atm-graft-python/src/query.rs
+++ b/crates/atm-graft-python/src/query.rs
@@ -67,7 +67,10 @@ impl PyGraftSession {
         // irrespective of the host session's chat identifier. The
         // session-chat filter is appropriate for bucket scans, but would hide
         // a bare-agent-addressed message that list just returned.
-        let exact_message_id = message_id.is_some();
+        let caller_chat_id = message_id
+            .is_none()
+            .then(|| self.caller.chat_id().cloned())
+            .flatten();
         if peek {
             PeekQuery::new(
                 home_dir,
@@ -85,14 +88,7 @@ impl PyGraftSession {
                 None,
             )
             .map_err(atm_error)
-            .map(|query| {
-                let query = if exact_message_id {
-                    query
-                } else {
-                    query.with_caller_chat_id(self.caller.chat_id().cloned())
-                };
-                ReadOperation::Peek(query)
-            })
+            .map(|query| ReadOperation::Peek(query.with_caller_chat_id(caller_chat_id)))
         } else {
             ReadQuery::new(
                 home_dir,
@@ -112,14 +108,14 @@ impl PyGraftSession {
             )
             .map_err(atm_error)
             .map(|query| {
-                let query = if exact_message_id {
+                ReadOperation::Read(
                     query
-                } else {
-                    query.with_caller_chat_id(self.caller.chat_id().cloned())
-                };
-                ReadOperation::Read(query.with_activity_observation(
-                    activity_observation_for_resolved_caller(self.caller.agent(), &team),
-                ))
+                        .with_caller_chat_id(caller_chat_id)
+                        .with_activity_observation(activity_observation_for_resolved_caller(
+                            self.caller.agent(),
+                            &team,
+                        )),
+                )
             })
         }
     }
--- a/crates/atm-graft-python/tests/test_cli_parity.py
+++ b/crates/atm-graft-python/tests/test_cli_parity.py
@@ -531,6 +531,44 @@ class CliParityTests(unittest.TestCase):
         self.assertEqual(read["count"], 1)
         self.assertTrue(read["mutation_applied"])
 
+    def test_native_peek_by_listed_id_reaches_bare_agent_message_from_chat(self) -> None:
+        sent = self._native(
+            self.second_native_tools,
+            "atm_send",
+            {
+                "to": f"{self.identity}@{self.team}",
+                "body": "native-exact-peek-regression",
+            },
+        )
+        message_id = sent["message_id"]
+        listed = self._native(self.native_tools, "atm_list", {"selection": "all"})
+        self.assertIn(message_id, {row["message_id"] for row in listed["rows"]})
+
+        peek = self._native(
+            self.native_tools,
+            "atm_read",
+            {"selection": "all", "message_id": message_id, "peek": True},
+        )
+        self.assertEqual(peek["count"], 1)
+        self.assertFalse(peek["mutation_applied"])
+
+    def test_cli_peek_by_listed_id_reaches_bare_agent_message_from_chat(self) -> None:
+        sent = self._native(
+            self.second_native_tools,
+            "atm_send",
+            {
+                "to": f"{self.identity}@{self.team}",
+                "body": "cli-exact-peek-regression",
+            },
+        )
+        message_id = sent["message_id"]
+        listed = self._native(self.native_tools, "atm_list", {"selection": "all"})
+        self.assertIn(message_id, {row["message_id"] for row in listed["rows"]})
+
+        peek = self._cli("peek", "--message-id", message_id, identity=self.identity)
+        self.assertEqual(peek["count"], 1)
+        self.assertFalse(peek["mutation_applied"])
+
 
 if __name__ == "__main__":
     unittest.main()
--- a/crates/atm-graft-python/tests/test_cli_parity.py
+++ b/crates/atm-graft-python/tests/test_cli_parity.py
@@ -552,18 +552,16 @@ class CliParityTests(unittest.TestCase):
         self.assertEqual(peek["count"], 1)
         self.assertFalse(peek["mutation_applied"])
 
-    def test_cli_peek_by_listed_id_reaches_bare_agent_message_from_chat(self) -> None:
+    def test_cli_peek_by_id_reaches_other_session_message_from_chat(self) -> None:
         sent = self._native(
             self.second_native_tools,
             "atm_send",
             {
-                "to": f"{self.identity}@{self.team}",
+                "to": f"{self.identity}:other-session@{self.team}",
                 "body": "cli-exact-peek-regression",
             },
         )
         message_id = sent["message_id"]
-        listed = self._native(self.native_tools, "atm_list", {"selection": "all"})
-        self.assertIn(message_id, {row["message_id"] for row in listed["rows"]})
 
         peek = self._cli("peek", "--message-id", message_id, identity=self.identity)
         self.assertEqual(peek["count"], 1)
--- a/crates/atm-graft-python/tests/test_cli_parity.py
+++ b/crates/atm-graft-python/tests/test_cli_parity.py
@@ -104,6 +104,10 @@ class _GeneratedParityFixture:
             "ATM_TEST_RUNTIME_HOME": str(home),
             "ATM_TEAM": self.team,
             "ATM_CHAT_ID": self.chat_id,
+            # Exercise the capability-authenticated loopback client even on
+            # Unix, where UDS is normally preferred. The regressions below
+            # require a real daemon and specifically cover TCP keep-alive.
+            "ATM_LOCAL_TRANSPORT": "tcp",
             "ATM_LOG_DIR": str(log_dir),
             "TMPDIR": str(temp_dir),
             "TMP": str(temp_dir),
@@ -461,6 +465,55 @@ class CliParityTests(unittest.TestCase):
             pending_ids = {row["message_id"] for row in pending["rows"]}
             self.assertNotIn(native_send["message_id"], pending_ids)
             self.assertNotIn(cli_send["message_id"], pending_ids)
+
+    def test_native_loopback_write_survives_idle_server_header_timeout(self) -> None:
+        # Warm this session's reqwest client, then exceed the daemon's default
+        # three-second HTTP/1 header timer. The next native request must open
+        # a fresh loopback connection rather than write to the stale pool.
+        self._native(self.native_tools, "atm_list", {"selection": "all"})
+        time.sleep(4)
+        result = self._native(
+            self.native_tools,
+            "atm_send",
+            {
+                "to": f"{self.second_identity}@{self.team}",
+                "body": "idle-loopback-regression",
+            },
+        )
+        self.assertIn("message_id", result)
+
+    def test_native_read_by_listed_id_marks_bare_agent_message_read(self) -> None:
+        sent = self._native(
+            self.second_native_tools,
+            "atm_send",
+            {
+                "to": f"{self.identity}@{self.team}",
+                "body": "native-exact-read-regression",
+            },
+        )
+        message_id = sent["message_id"]
+        listed = self._native(self.native_tools, "atm_list", {"selection": "all"})
+        self.assertIn(message_id, {row["message_id"] for row in listed["rows"]})
+
+        read = self._native(
+            self.native_tools,
+            "atm_read",
+            {"selection": "all", "message_id": message_id},
+        )
+        self.assertEqual(read["count"], 1)
+        self.assertTrue(read["mutation_applied"])
+
+        # Read acceptance is intentionally asynchronous; poll only the
+        # disposable daemon's list projection for the bounded handoff.
+        deadline = time.monotonic() + 3
+        while time.monotonic() < deadline:
+            current = self._native(self.native_tools, "atm_list", {"selection": "all"})
+            matching = [row for row in current["rows"] if row["message_id"] == message_id]
+            if matching and matching[0]["read"]:
+                break
+            time.sleep(0.05)
+        else:
+            self.fail("native exact-ID read was not visible after the bounded handoff")
 
 
 if __name__ == "__main__":
--- a/crates/atm-graft-python/tests/test_cli_parity.py
+++ b/crates/atm-graft-python/tests/test_cli_parity.py
@@ -104,10 +104,6 @@ class _GeneratedParityFixture:
             "ATM_TEST_RUNTIME_HOME": str(home),
             "ATM_TEAM": self.team,
             "ATM_CHAT_ID": self.chat_id,
-            # Exercise the capability-authenticated loopback client even on
-            # Unix, where UDS is normally preferred. The regressions below
-            # require a real daemon and specifically cover TCP keep-alive.
-            "ATM_LOCAL_TRANSPORT": "tcp",
             "ATM_LOG_DIR": str(log_dir),
             "TMPDIR": str(temp_dir),
             "TMP": str(temp_dir),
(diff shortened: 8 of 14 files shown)


randlee:
## Final Quality Report — QA-1 (qa-1299-r1-1788799419)

**PR:** #1299 (`fix/native-loopback-157` → `develop`)
**Head:** `c6a2ed1586154dd92e30f7bf88ca49403644e4f5` (one commit unique to this branch, `63e0c2b33`, on top of `develop`'s merge-base `8845f3b39`; unrelated colima-docs content that arrived via a `develop` merge was excluded from scope)
**Responds to:** GitHub issues #1297 (stale-pooled keep-alive connection causes a false `ATM_DAEMON_MAY_HAVE_EXECUTED` after ~3s server-side idle) and #1298 (native `atm_read` by exact `message_id` wrongly applies the session's chat_id filter, returning `count=0`)
**Verdict:** ❌ **FAIL 1/2/4** (1 Blocking, 2 Important, 4 Minor)
**Deliverable completion:** issue #1297 — fixed only on the non-default TCP-loopback client, **not** on the default Unix-domain-socket client (partial). Issue #1298 — fixed and rigorously proven for the native `atm-graft-python` path, **not** fixed for the CLI's own equivalent path, which carries the identical latent bug (partial). **2/4 named sub-fixes fully closed, 2/4 partially closed (50%)**.

Every reviewer finding below was independently re-verified by quality-mgr directly against the diff (`git show`/direct file reads), not accepted on a reviewer's word alone.

### Reviewer results (6/6 returned)

| Reviewer | Verdict | Result |
|---|---|---|
| req-qa | FAIL (1 Blocking) | Issue #1298 fix confirmed sound and correctly scoped (no cross-mailbox/team widening); issue #1297 fix confirmed incomplete — applies only to the TCP-loopback client, not UDS (the actual default transport) |
| arch-qa | PASS (1 Important) | Same UDS-gap finding as req-qa, rated Important rather than Blocking; confirms query.rs's chat-id skip never widens authorization scope; notes CLI/graft asymmetry (see RBQA-F001) as informational |
| rust-qa-agent | PASS (1 Minor) | fmt/clippy/full workspace test suite (incl. `atm-daemon`) clean; version-sync and line-count gates clean; **built and ran both new regression tests against real daemon binaries, and against a baseline pre-fix checkout — both fail-then-pass exactly as expected**, proving the fixes are real, not vacuous |
| ruthless-boundary-qa | FAIL (1 Critical, 1 Minor) | New Critical finding: CLI's `atm read --message-id` has the identical #1298 defect, unfixed; Minor: duplicate `ATM_LOCAL_TRANSPORT` parser |
| rust-best-practices-agent | PASS (1 Important, informational) | Duplicate `ATM_LOCAL_TRANSPORT` parsing has already drifted behaviorally (non-unix arm silently ignores an invalid value, unlike the canonical resolver) |
| rust-service-hardening-agent | FAIL (1 Important, 2 Minor) | New Important finding: disabling the TCP-loopback pool entirely (not just shortening its idle timeout) removes keep-alive reuse for legitimate long-lived sessions (e.g. Windows `PyGraftSession`), unvalidated under load; 2 Minor observability/consistency nits |

### Blocking finding — independently reproduced by quality-mgr

**QA-B1 — The #1297 fix only covers the non-default TCP-loopback client; the default Unix-domain-socket client keeps the same unbounded idle pool that caused the original defect, and the new regression test is structured so it cannot catch this.** (req-qa ATM-QA-001, arch-qa ARCH-001 — independently confirmed by quality-mgr)

- `grep -rn "pool_max_idle_per_host" crates` returns exactly **one** match in the whole workspace: `crates/atm-http-runtime/src/client.rs:488`, inside `build_loopback_reqwest_client` (the TCP connector).
- `build_unix_socket_reqwest_client` (`crates/atm-http-runtime/src/client_tail.rs:106-118`) builds its `reqwest::Client` with no idle-pool override at all — confirmed by direct read, no such call present.
- `preferred_local_client` (`crates/atm-http-runtime/src/client.rs:367-417`) selects **UDS by default** on Unix — the TCP-loopback path is only reached via the explicit `ATM_LOCAL_TRANSPORT=tcp` diagnostic override or a root-owned runtime directory (`uid == 0`) — confirmed by direct read of the full function body. UDS is therefore the transport actually used by native `atm`/`atm-graft` callers in ordinary macOS/Linux operation, the exact scenario issue #1297 was filed against.
- Both connectors are served by the identical `header_read_timeout`/keep-alive policy on the daemon side (arch-qa cites `crates/atm-http-runtime/src/http1_server.rs`'s `serve_loopback_http1` vs `serve_unix_http1`, same policy, lines 160-201 vs 261-291) — so the same stale-pooled-connection race is mechanically reachable over UDS, not just TCP.
- The new regression test, `test_native_loopback_write_survives_idle_server_header_timeout` (`crates/atm-graft-python/tests/test_cli_parity.py:107-110,469-483`), deliberately sets `ATM_LOCAL_TRANSPORT=tcp` in its fixture environment — so CI proves the TCP path is fixed but structurally cannot detect a recurrence on the default UDS path.
- The code comment at `client.rs:483-487` ("this is deliberately local-only: peer clients retain their pool") overstates the fix's actual coverage — UDS is likewise a local, non-peer client, and it was left out.

This is Blocking because it means the PR does not actually close #1297 for the common case — only for an opt-in diagnostic mode. **Required correction:** apply the same (or an equivalent bounded) idle-pool policy to `build_unix_socket_reqwest_client`, or provide a written, verified justification for why UDS is exempt from the same race (e.g. a real platform difference in half-close detection — not asserted or evidenced anywhere in this PR), plus a CI-exercised regression test against the *default* UDS transport (not only the `ATM_LOCAL_TRANSPORT=tcp` mode).

### Important findings (non-blocking)

**QA-I1 — The CLI's own read path (`atm read --message-id`) has the identical #1298 defect this PR fixes only in native `atm-graft-python`.** (ruthless-boundary-qa RBQA-F001, rated Critical there — I side with Important: this is a pre-existing bug outside this PR's diff, not a regression it introduces, but it directly undermines the fix's own stated justification.) Independently confirmed by quality-mgr via direct read:
- `crates/atm/src/commands/read.rs::build_query` (lines 108-153) unconditionally calls `.with_caller_chat_id(caller_context.caller_chat_id)`, with no branch on `self.message_id`.
- `select_classified_mailbox_candidates`'s exact-message-id branch (`crates/atm-core/src/read/selection.rs:144-151`) still applies `filters::matches_participant_filter(message, request.participant_filter.as_ref())`, and `matches_participant_filter` (`crates/atm-core/src/read/filters.rs:55-73`) requires an exact chat_id match — so a chat-scoped CLI caller (`--chat-id` set, or an ambient chat-qualified identity) doing `atm read --message-id X` for a message addressed to the bare-agent identity will not find it, reproducing #1298's exact symptom in the CLI.
- The new graft-python fix's own justification comment (`crates/atm-graft-python/src/query.rs:66-70`) asserts the CLI already does the right thing here — that assertion is false against this diff.
Recommend: give `ReadCommand::build_query` the same bypass (skip `.with_caller_chat_id(...)` when `message_id.is_some()`), or better, move the rule into `ReadQuery`/`PeekQuery` construction itself so every caller inherits it from one place. At minimum, correct the false comment and file a tracked follow-up before merge if the CLI fix itself is deferred.

**QA-I2 — Disabling the TCP-loopback pool entirely (`pool_max_idle_per_host(0)`), rather than bounding its idle timeout, removes connection reuse for legitimate long-lived callers on the only platform where this connector is the sole local transport.** (rust-service-hardening-agent RSH-002, confirmed plausible by quality-mgr via code read — not independently load-tested) Windows always uses this TCP-loopback connector (`preferred_local_client`'s `#[cfg(not(unix))]` arm, `client.rs:413-416`), and the code's own doc comment names a long-lived embedded caller (`PyGraftSession`, `client.rs:426-435`) that will now pay a fresh TCP handshake on every single send/read for the life of that session, not just a one-off CLI call. The correctness rationale (stale sockets stalling silently) is sound, but the performance tradeoff is currently unvalidated under sustained load. Recommend a throughput check or a reconnect-volume counter as a follow-up, not necessarily a merge blocker on its own.

### Minor findings (non-blocking)

- **QA-M1 (rust-best-practices-agent RBP-F001 / ruthless-boundary-qa RBQA-F002 / rust-qa-agent QA-001 / arch-qa's coupling note)** — `preferred_local_client`'s new `ATM_LOCAL_TRANSPORT` parsing (`client.rs:373-386`) duplicates the pre-existing, unit-tested `atm_daemon_client::local_transport::LocalDaemonTransport::resolve` (`crates/atm-daemon-client/src/local_transport.rs:16-40`), with matching validation text but no shared implementation or test tying them together. Confirmed by quality-mgr this is architecturally forced, not careless: `atm-daemon-client` depends on `atm-http-runtime` (`crates/atm-daemon-client/Cargo.toml:14`), so the dependency cannot run the other way — `atm-http-runtime` cannot call the existing resolver without a crate cycle. The two copies have already drifted: the new copy's `#[cfg(not(unix))]` arm (unchanged by this PR, `client.rs:413-416`) doesn't read `ATM_LOCAL_TRANSPORT` at all, silently ignoring an invalid value on Windows, unlike the canonical resolver which validates on every platform. Recommend a boundary/architecture test pinning the two validation strings together until Phase AM's legacy-daemon deletion removes the duplication on its own — not a rewrite mandate given the Legacy Daemon Exemption.
- **RSH-001** — `ATM_LOCAL_TRANSPORT=""` (explicitly empty) is silently treated as unset/default with no distinguishing log; mirrors an existing convention in the legacy resolver, flagged only because the task asked about completeness on this exact axis.
- **RSH-003** — The new root-owned-runtime detection's `std::fs::metadata` call runs inline (not via `spawn_blocking`), inconsistent with this same file's own established pattern elsewhere; negligible practical impact (single local `stat()`).
- **Version/doc hygiene** — all confirmed clean: `python3 .just/check_version_sync.py` (1.5.7 across `Cargo.toml`, `Cargo.lock`, both `pyproject.toml`s, the winget manifest) and `python3 .just/check_line_counts.py` (RULE-003) both pass; `docs/requirements.md:4073` already documented the `ATM_LOCAL_TRANSPORT=tcp` mode this PR newly implements, so no scope-creep against undocumented behavior.

### The regression tests are real, not vacuous — independently proven by rust-qa-agent

rust-qa-agent built the `atm_graft_python` native extension and a disposable `atm`/`atm-daemon` pair from this worktree, then ran both new tests (`test_native_loopback_write_survives_idle_server_header_timeout`, `test_native_read_by_listed_id_marks_bare_agent_message_read`) via `ATM_CLI_PARITY_CI=1` — both PASS, twice, no flakiness. It then built a **second** checkout at the pre-fix baseline (`f20441cc7`, the last commit before this PR) and ran the identical tests against it: both **FAIL** exactly as the two issues describe (`KeyError('result')` / MAY_HAVE_EXECUTED fallback for #1297; `count == 0` instead of `1` for #1298). This is the single strongest piece of evidence in this round — it proves the fixes are real for the paths they cover, which is also what makes QA-B1's scope gap (UDS untested) and QA-I1's scope gap (CLI untested) stand out as real omissions rather than reviewer pedantry.

### Merge Readiness
- Status: **not ready**. 1 Blocking finding (QA-B1) gates merge — the PR does not actually close #1297 for the default (UDS) transport, only for an opt-in diagnostic TCP mode. Per policy, fix rounds route to **arch-ctm**.
- QA-I1 (CLI parity gap) should be fixed or explicitly deferred-with-tracking before merge — it directly contradicts the fix's own justification comment.
- `mergeable: MERGEABLE`, `mergeStateStatus: BLOCKED` only on pending/required CI (not a real conflict). Branch is not part of a `gh stack`.

---
*Posted by quality-mgr. Zero `.triage/`, `.sprints/`, or branch/worktree writes made by quality-mgr.*

randlee:
Validation outputs for b468ae96d:

- cargo fmt --check: PASS
- cargo clippy --workspace --all-targets -- -D warnings: PASS
- cargo test --workspace --quiet: PASS (workspace suite, including atm-daemon)
- python3 .just/check_version_sync.py: PASS (1.5.7 lockstep)
- python3 .just/check_line_counts.py: PASS
- cargo build -p agent-team-mail -p atm-daemon -p atm-graft-python: PASS
- generated disposable-daemon native/CLI parity: PASS (4 tests; default UDS idle write and CLI exact-ID coverage included)
- python3 scripts/test_atm_graft_python.py: PASS (wheel import and receiver ownership/reclaim checks)

No post-QA push will be made unless QA identifies a required correction.

randlee:
## Final Quality Report — QA-2 (qa-1299-r2-1788800848)

**PR:** #1299 (`fix/native-loopback-157` → `develop`)
**Head:** `b468ae96d2efd481787f94735a9554910b7ade9b` (R2, one commit `fix: bound local HTTP idle pools` on top of QA-1's `c6a2ed158`)
**Responds to:** GitHub issue #1297 (stale-pooled keep-alive connection causes a false `ATM_DAEMON_MAY_HAVE_EXECUTED`) and #1298 (native `atm_read` by exact `message_id` wrongly applies the chat_id filter); QA-1 verdict FAIL 1/2/4 (https://github.com/randlee/atm-core/pull/1299#issuecomment-5573686819)
**Verdict:** ❌ **FAIL 0/1/1** (0 Blocking, 1 Important — newly surfaced, confirmed pre-existing and non-blocking, 1 Minor — accepted documented debt)
**Deliverable completion (against issues #1297/#1298's actual stated scope):** **4/4 (100%)** — all QA-1 carried-forward findings are closed with real evidence, including the two most important ones (the default-UDS transport gap and the CLI parity gap).

Every reviewer finding below was independently re-verified by quality-mgr directly against the diff (`git diff c6a2ed158..b468ae96d`, direct file reads), not accepted on a reviewer's word alone.

### Reviewer results (6/6 returned)

| Reviewer | Verdict | Result |
|---|---|---|
| req-qa | FAIL (1 new Blocking, disposition below) | QA-B1 and QA-I1 both confirmed closed with concrete evidence; new finding ATM-QA-001 (REQ-CORE-TRANSPORT-001 diagnostic-mode contract no longer honored by the production path after the R1 parser was removed) |
| arch-qa | PASS (0/0) | QA-B1 and the coupling note both fully resolved; no speculative/unused code found anywhere in the R2 diff |
| rust-qa-agent | PASS (0/0) | fmt/clippy/full workspace test suite (incl. `atm-daemon`) clean; **built the native extension + a disposable daemon pair and ran all 4 relevant tests — all pass; additionally fault-injected the regression by reverting just the UDS fix and confirmed the test correctly fails, then restored and confirmed it passes again** — the strongest possible proof the fix and its test are both real |
| ruthless-boundary-qa | PASS (0/0, findings dispositioned "fixed") | RBQA-F001 and RBQA-F002 both confirmed closed; no dead code remnants |
| rust-best-practices-agent | PASS (0/0) | RBP-F001 confirmed fully resolved — the duplicate parser is gone, not patched in place |
| rust-service-hardening-agent | PASS (1 Minor) | RSH-002 and RSH-001 fully resolved; RSH-003 carried forward as an accepted, now-documented minor (no code change needed) |

### QA-B1 (Blocking at QA-1) — confirmed closed by 4 independent reviewers plus quality-mgr

Both `build_loopback_reqwest_client` (`client.rs:490-494`) and `build_unix_socket_reqwest_client` (`client_tail.rs:106-116`) now route through one shared `local_reqwest_client_builder()` (`client.rs:479-484`), which sets `.pool_idle_timeout(LOCAL_CLIENT_IDLE_POOL_TIMEOUT)` (`= Duration::from_secs(2)`, comfortably below the daemon's 3s header-read timer) — a single owner, bounded-idle policy, not a full pool disable and not two independently-tuned copies. `grep -rn "pool_max_idle_per_host\|pool_idle_timeout" crates` confirms exactly one idle-pool policy definition in the whole workspace.

The regression test now genuinely exercises the **default** path: `test_native_default_uds_write_survives_idle_server_header_timeout` (renamed from the QA-1 version) asserts `ATM_LOCAL_TRANSPORT` is **not** set in the fixture environment, then reproduces the idle-then-write scenario. rust-qa-agent went further than any other round this PR has seen: it built the real native extension and a disposable daemon pair, confirmed via `lsof` that the session genuinely holds a Unix-domain socket (zero inet sockets) during the test, then **deliberately reverted only the UDS half of the fix**, rebuilt, and reran the test — it failed exactly as #1297 describes (`KeyError('result')`) — before restoring and reconfirming a clean pass. This is conclusive: the fix and the test are both real and load-bearing on the actual default transport, closing the exact gap QA-1 found Blocking.

### QA-I1 (Important at QA-1) — confirmed closed by ruthless-boundary-qa, rust-qa-agent, req-qa, and quality-mgr

`crates/atm/src/commands/read.rs::build_query` now skips `.with_caller_chat_id(...)` when `self.message_id.is_some()` (with an inline comment explaining the rationale), backed by a new unit test (`exact_message_id_does_not_apply_session_chat_scope`) and a new real-daemon integration test (`test_cli_read_by_listed_id_reaches_bare_agent_message_from_chat`) that sends via native graft from one identity and reads it back via the **CLI** with a chat_id set on the caller — proving the CLI path, not just the native path, now correctly reaches a bare-agent-addressed message by exact ID. The `atm-graft-python/src/query.rs` comment that previously falsely claimed CLI parity has been corrected to state the true, symmetric rationale.

### QA-I2 / RSH-001 (Important / Minor at QA-1) — both confirmed closed

QA-I2 (full pool disable, no keep-alive reuse) is resolved by the same bounded 2s `pool_idle_timeout` described above — reuse is retained within a burst, not eliminated. RSH-001 (empty-vs-unset `ATM_LOCAL_TRANSPORT` indistinguishable) is resolved: `local_daemon_transport()` (`crates/atm-daemon-client/src/local_transport.rs:54-68`) now logs an `override_state` field (`unset`/`explicitly_empty`/`explicit`) alongside the resolved transport.

### RBP-F001 / RBQA-F002 (duplicate parser) — confirmed fully removed, not just patched

The `ATM_LOCAL_TRANSPORT` parser R1 added to `preferred_local_client` is gone entirely — confirmed by every reviewer that checked (arch-qa, ruthless-boundary-qa, rust-best-practices-agent, rust-qa-agent, req-qa) via direct grep/read: zero references to `ATM_LOCAL_TRANSPORT` or `std::env::var` remain in `client.rs`. `preferred_local_client` now selects UDS-vs-loopback purely via the pre-existing root-uid ownership check, matching baseline shape. The sole surviving parser is the pre-existing, unit-tested `atm-daemon-client::local_transport::LocalDaemonTransport::resolve`, extended only by the RSH-001 observability field above.

### RSH-003 (Minor, carried forward, unchanged) — accepted as documented debt

`runtime_directory_is_root_owned()` still performs an inline `std::fs::metadata()` call rather than routing through `spawn_blocking`, inconsistent with this file's own convention elsewhere. R2 added a doc comment directly justifying this ("a single local `stat()` syscall, no daemon I/O, executed once per client construction, not per-request"). rust-service-hardening-agent accepts this as documented, non-blocking technical debt — no code change required.

### New finding this round — independently adjudicated by quality-mgr as Important, not Blocking

**ATM-QA-001 (req-qa, rated Blocking there; quality-mgr downgrades to Important based on independent history verification below)** — `docs/requirements.md:4073-4075` (`REQ-CORE-TRANSPORT-001`) states: *"Unix clients select UDS by default. `ATM_LOCAL_TRANSPORT=tcp` is the explicit, observable loopback-TCP parity/diagnostic mode..."* Since R2 correctly removed R1's duplicate parser from `preferred_local_client` (the actual production connector-selection function), that function — and therefore the real CLI/graft write and read paths — no longer reads `ATM_LOCAL_TRANSPORT` at all. The only remaining consumers of the env var are `atm-daemon-client`'s `#[deprecated]` legacy synchronous path and `GraftClient::local_transport_label()` (surfaced as `endpoint_kind` in AW.4 recovery observability), which independently re-reads the env var and can therefore report a transport label that disagrees with what `preferred_local_client` actually selected.

Quality-mgr independently verified via `git show f20441cc7:crates/atm-http-runtime/src/client.rs` (the commit immediately before PR #1299 existed) that **this exact condition already existed at baseline, before this PR touched anything**: `preferred_local_client` never read `ATM_LOCAL_TRANSPORT` at baseline either, and `local_transport_label()`/`recovery_endpoint_kind()` already had the identical env-var-vs-actual-connector disconnect, unchanged by this PR's diff in either R1 or R2. R1 temporarily (and at unacceptable cost — a duplicated, already-drifted parser) made the diagnostic mode real for the production path; R2 correctly reverted that per explicit direction that transport selection must live in exactly one place. Net effect: this PR restores, rather than introduces, the pre-existing condition.

**Disposition: real finding, confirmed, but not a regression introduced by this PR and not a blocker for merging the #1297/#1298 fixes.** Recommend: file a tracked follow-up issue to either (a) properly thread `ATM_LOCAL_TRANSPORT` into the production path without duplicating the parser (e.g. a narrow shared resolution point both crates can depend on, breaking the current one-directional dependency constraint), or (b) narrow `REQ-CORE-TRANSPORT-001`'s wording to the legacy path and re-derive `recovery_endpoint_kind()`/`local_transport_label()` from the actual connector that was built rather than an independent env-var re-read.

### Checks confirmed directly by quality-mgr

`cargo fmt --all --check`, `cargo clippy --all-targets --all-features -- -D warnings`, `cargo test --workspace` (0 failures) and `cargo test -p atm-daemon` explicitly (`version_string_matches_the_binary_contract` passes), `python3 .just/check_version_sync.py` (workspace_version=1.5.7, all pins aligned — unchanged from QA-1, R2 touches no version file), `python3 .just/check_line_counts.py` (no new RULE-003 violations). `gh pr view 1299`: `mergeable: MERGEABLE`, `mergeStateStatus: BLOCKED` only on pending/required CI (not a conflict); not part of a `gh stack`.

### Merge Readiness
- Status: **ready** for this PR's actual purpose. 0 Blocking findings. The one new Important finding (ATM-QA-001) is independently confirmed pre-existing (verified against the pre-PR baseline commit `f20441cc7`) and should not gate this hotfix — recommend a separate tracked follow-up issue rather than another fix round on this branch.
- All five QA-1 carried-forward findings (QA-B1, QA-I1, QA-I2, RBP-F001/RBQA-F002, RSH-001) are closed with concrete, independently-verified evidence, including a fault-injection proof (rust-qa-agent) that the core fix and its test are both real.
- RSH-003 remains open as accepted, documented, non-blocking technical debt.

---
*Posted by quality-mgr. Zero `.triage/`, `.sprints/`, or branch/worktree writes made by quality-mgr.*

randlee:
fenix scope ruling on QA-2 (FAIL 0/1/1): ATM-QA-001 is verified pre-existing at baseline f20441cc7 (preferred_local_client never read ATM_LOCAL_TRANSPORT; label/connector disconnect predates this PR) and is out of scope for #1297/#1298. Tracked as #1303 together with RSH-003. Rand's rulings governing it: transport selection lives once, and no code is added that is not necessary and used consistently, so the R1 parser is not coming back. With that ruling the live count for this branch is 0/0/0; merge gate is CI green.

randlee:
R3 validation for 0122a9c9f:

- python3 .just/run_lint.py function-length: PASS (build_tool_read_query is 77 production lines; cap 80)
- python3 .just/run_lint.py lines: PASS
- python3 .just/run_lint.py boundaries: PASS
- python3 .just/run_lint.py version: PASS
- cargo fmt --check: PASS
- cargo clippy --workspace --all-targets -- -D warnings: PASS
- cargo test -p atm-graft-python: PASS (39 tests)
- Rebuilt the debug binding at 0122a9c9f, then ran the two disposable real-daemon parity regressions: PASS (default UDS idle write; CLI exact-ID read with fixture chat context).

The all-target invocation shown in the assignment is unsupported by run_lint.py, so the four required targets were run individually. The commit is net -4 lines and introduces no function or abstraction. Holding this branch for QA-3.

randlee:
## Final Quality Report — QA-3 (qa-1299-r3-1788802096)

**PR:** #1299 (`fix/native-loopback-157` → `develop`)
**Head:** `0122a9c9f` (R3, one file `crates/atm-graft-python/src/query.rs`, +12/-16, on top of QA-2's `b468ae96d`)
**Stated purpose:** clear CI's RULE-002 function-length lint failure on `build_tool_read_query` (81 lines) by computing the exact-ID/chat-scope decision once before the Peek/Read split — claimed by its commit message and by the dispatching task to be a **pure refactor with no behavior change**.
**Verdict:** ❌ **FAIL 1/2/1** (1 Blocking, 2 Important, 1 Minor)
**Deliverable completion:** the RULE-002 lint fix itself is real and clean, but the commit is **not** behavior-preserving as claimed — it silently widens `atm_read(peek=True, message_id=...)`'s chat-scope matching, with zero test coverage of the new behavior.

Quality-mgr's own direct code analysis ahead of dispatch suspected this commit was not truly behavior-preserving; all three reviewers independently confirmed a real change, converged on the same precise scope after correcting my own overbroad initial premise, and rust-qa-agent went further and **proved it empirically** by executing the real production selection code, not just reading it.

### Reviewer results (3/3 returned)

| Reviewer | Verdict | Result |
|---|---|---|
| req-qa | FAIL (2 Important, 1 Minor) | Confirms the widening, scoped correctly to Peek only (not Read); finds an additional fresh CLI-vs-native-graft parity gap for `atm peek --message-id`; notes my dispatch premise incorrectly assumed both `ReadQuery::new` and `PeekQuery::new` share the same restrictive default |
| arch-qa | PASS (1 Important) | Confirms the same Peek-only widening via direct code read; confirms no architectural/boundary/RULE-001..013 violation and no ownership-boundary impact; routes the functional-scope call to req-qa/rust-qa-agent |
| rust-qa-agent | FAIL (1 Critical → adjudicated Blocking, 1 Important) | **Empirically proved** the widening by executing two temporary tests against the real production types (`PeekQuery::new`, `with_caller_chat_id`, `matches_participant_filter`), then reverted them cleanly; identifies the root cause as a pre-existing contract mismatch in `with_caller_chat_id` itself |

### Blocking finding — independently reproduced by quality-mgr, empirically proven by rust-qa-agent

**QA-3-B1 — R3's refactor is not behavior-preserving: native `atm_read(peek=True, message_id=...)` now matches a message addressed to *any* chat-qualified session of the caller's own agent identity, not just the bare-agent identity, and this shipped with zero test coverage under a commit that explicitly claims no behavior change.**

`PeekQuery::new` (`crates/atm-core/src/read/request.rs:218-257`) explicitly sets `mailbox.participant_filter = Some(MessageParticipantFilter{agent, chat_id: None, direction: To})` at construction — a bare-agent-only match. At QA-2's head (`b468ae96d`), the exact-message-id branch in `query.rs` skipped calling `with_caller_chat_id` entirely, leaving this restrictive default in place. At R3's head (`0122a9c9f`), the code now computes `caller_chat_id = message_id.is_none().then(|| self.caller.chat_id().cloned()).flatten()` (which is `None` for an exact-ID lookup) and **always** calls `.with_caller_chat_id(caller_chat_id)`. `with_caller_chat_id(None)` (`request.rs:267-278`) sets `participant_filter = None`, which `matches_participant_filter` (`crates/atm-core/src/read/filters.rs:55-70`) treats as "match everything" — a strictly wider match than the constructor's own default.

rust-qa-agent settled this empirically, not just by reading code: it added two temporary `cargo test` cases directly exercising `PeekQuery::new`/`with_caller_chat_id`/`matches_participant_filter` against a synthetic message with `destination_chat_id = Some("chat-other")`, confirmed the old code path did **not** match it and the new code path **does**, then cleanly reverted the temporary tests (`git status --short` confirms the worktree is clean, only the pre-existing untracked build artifact remains). **`ReadQuery` (the non-peek path, and the actual subject of issue #1298) is unaffected** — `ReadQuery::new` never overrode `participant_filter` in the first place, so it already defaulted to `None` before and after R3; quality-mgr independently confirmed this by direct read of `request.rs:333-369`.

This is Blocking because: (1) it is a real, empirically-proven production-code behavior change in a file this PR's own diff modifies; (2) it is falsely characterized as a pure refactor with no behavior change, in both the commit and the dispatching task description; (3) it ships with zero regression test coverage of either the old or the new semantics for Peek; (4) `filters.rs`'s own design comment states "a bare agent, and each chat-qualified instance of that agent, are distinct inbox identities" — this widening blurs exactly that invariant for exact-ID peeks.

**Root cause (rust-qa-agent QA-002, folded into this finding):** `with_caller_chat_id`'s own doc comment states `None` "selects only the bare-agent identity," but its implementation actually clears `participant_filter` to `None` (match-everything) — the implementation contradicts its own documented contract. R3 didn't introduce this latent bug; it newly exercises it via an unconditional call on a path that previously never invoked this function for exact-ID Peek lookups.

**Required correction:** either (a) restore the prior skip-branch behavior for Peek's exact-ID case (leave `participant_filter` at `PeekQuery::new`'s own default rather than clearing it), or (b) if unscoping exact-ID peek is the intended design (the corrected code comment at `query.rs:66-69` argues this), fix `with_caller_chat_id` to match its own contract, make the widening an explicit, documented, tested decision — not a side effect of a lint fix — and align `crates/atm/src/commands/peek.rs` (see QA-3-I2 below) to the same semantics. Route to arch-ctm.

### Important findings (non-blocking on their own, but tied to the same root cause)

**QA-3-I1 (rust-qa-agent QA-002)** — see root-cause paragraph above: `with_caller_chat_id(None)`'s implementation contradicts its own doc comment. This is pre-existing (the file is outside R3's diff) but newly reachable/live because of R3; fixing QA-3-B1 properly requires resolving this contract mismatch, not papering over it at the call site.

**QA-3-I2 (req-qa ATM-QA-102)** — R3 reintroduces a CLI-vs-native-graft parity gap, confined to Peek: `crates/atm/src/commands/peek.rs` (lines 94-134) never calls `with_caller_chat_id` at all and always keeps the restrictive bare-agent-only default, so CLI's `atm peek --message-id` and native graft's `atm_read(peek=True, message_id=...)` now disagree on whether a message addressed to a different chat-qualified session of the same agent is found — the opposite-direction version of the parity gap QA-1/QA-2 fixed for Read. Whichever semantics is chosen for QA-3-B1's fix must be applied identically to both entry points, with a parity test covering both.

### Minor finding

**QA-3-M1 (req-qa ATM-QA-103)** — the dispatching task's own verification premise incorrectly assumed both `ReadQuery::new` and `PeekQuery::new` share the same restrictive construction default; only `PeekQuery::new` does. Noted so a follow-up fix round is correctly scoped to Peek only and doesn't waste effort on a Read-path fix for a defect that doesn't exist there.

### Checks confirmed directly by quality-mgr

`python3 .just/run_lint.py function-length` passes clean (RULE-002 resolved as intended). `cargo fmt --all --check` and `cargo clippy -p atm-graft-python --all-targets --all-features -- -D warnings` both clean; rust-qa-agent additionally reran full-workspace `clippy`/`cargo test --workspace`/`cargo test -p atm-daemon` fresh — all clean, 0 failures (the temporary empirical-proof tests were added and reverted within that same clean run). No version file touched by R3; version lockstep at 1.5.7 unchanged from QA-2. `gh pr view 1299`: `mergeable: MERGEABLE`, `mergeStateStatus: BLOCKED` only on pending CI; not part of a `gh stack`.

### Merge Readiness
- Status: **not ready**. 1 Blocking finding (QA-3-B1) gates merge — a real, empirically-proven, untested behavior change shipped under a false "no behavior change" claim. Per policy, route the fix round to **arch-ctm**.
- The two Important findings (root-cause contract mismatch, CLI/graft peek parity) should be resolved together with QA-3-B1's fix, not separately, since they share one root cause and one correct resolution.
- The prior QA-2 findings (QA-B1, QA-I1, QA-I2, RBP-F001/RBQA-F002, RSH-001, RSH-003) remain closed/accepted; this round only concerns the new R3 delta.

---
*Posted by quality-mgr. Zero `.triage/`, `.sprints/`, or branch/worktree writes made by quality-mgr.*

randlee:
fenix ruling on QA-3 (FAIL 1/2/1): an exact message id inside the caller's own mailbox is resolved by ownership only, never by chat scope, identically for read and peek, CLI and native. This is what requirements.md already says for --message-id ('resolve that exact message when present') and it is the semantics that fixed #1298 for read at R2. R3's peek behaviour is therefore the correct behaviour; the defects are the false no-behaviour-change claim, the CLI peek parity gap, the with_caller_chat_id doc comment contradicting its implementation, and the missing test. R4 dispatched to arch-ctm: CLI peek exact-ID bypass, doc comments corrected (implementation unchanged), tests for all four paths, PR body corrected. Making with_caller_chat_id(None) restrictive is rejected: chat-qualified hermes sessions would miss their own chat-addressed messages by id and #1298 would reopen.

randlee:
R4 validation for c32de7093:

- python3 .just/run_lint.py function-length: PASS
- python3 .just/run_lint.py lines: PASS
- python3 .just/run_lint.py boundaries: PASS
- python3 .just/run_lint.py version: PASS
- cargo fmt --check: PASS
- cargo clippy --workspace --all-targets -- -D warnings: PASS
- cargo test --workspace --quiet: PASS, including atm-daemon
- Rebuilt the debug binding at c32de7093 and ran the disposable real-daemon parity suite: PASS, 6 tests. This covers default-UDS idle write plus native and CLI exact-ID read and peek under fixture chat context.

The PR body now explicitly records the intentional R3 exact-ID peek behavior change, the ownership-only ruling, and R4s CLI parity/documentation coverage. Branch held for QA-4.

randlee:
## Final Quality Report — QA-4 (qa-1299-r4-1788803109)

**PR:** #1299 (`fix/native-loopback-157` → `develop`)
**Head:** `c32de7093` (R4, 4 files +115/-3, on top of QA-3's `0122a9c9f`)
**Task:** align CLI peek + doc comments + tests to fenix's ruling: *"an exact message id inside the caller's own mailbox resolves by ownership only, never by chat scope, identically for read and peek, CLI and native."*
**Verdict:** ❌ **FAIL 0/3/0** (0 Blocking, 3 Important, 0 Minor)
**Deliverable completion:** the ruling itself is correctly implemented and empirically proven across all four exact-ID paths (CLI read, CLI peek, native read, native peek) — but two of R4's own explicitly-required deliverables (test proof, doc-comment correction) are incomplete, and reviewing this closely surfaced one pre-existing, out-of-scope defect.

### Reviewer results (3/3 returned)

| Reviewer | Verdict | Result |
|---|---|---|
| req-qa | FAIL (1 Blocking, 1 Important) | Ruling correctly implemented, ownership boundary untouched; doc-comment defect rated Blocking because it falsifies the PR body's own "Closed" disposition claim |
| arch-qa | PASS (1 Minor) | No architectural/boundary/RULE-001..013 violation; ownership gate (`ensure_owner_only_read_target`) confirmed untouched; no speculative/unused code; doc-comment defect rated Minor (doc-only, no lint gate enforces it) |
| rust-qa-agent | FAIL (1 Critical, 1 Important, 2 Minor) | **Empirically proved** (fault injection) the ruling is correctly implemented for both bare-agent and chat-qualified destinations on all four paths; found the new CLI-peek test doesn't actually prove the scenario it's meant to guard; found a separate, pre-existing, out-of-scope chat-blindness bug in `atm peek` bucket scans; confirmed the same doc-comment defects |

### Finding 1 (Important) — R4's own ruling is correctly implemented and empirically proven (no defect — recorded for completeness)

All three reviewers independently confirmed, and rust-qa-agent additionally proved via fault injection (reverting `peek.rs`'s fix and re-running against a real disposable CLI/daemon pair, then restoring), that CLI read, CLI peek, native read, and native peek now all resolve an exact-ID lookup by mailbox ownership alone, for both bare-agent-addressed and chat-qualified-addressed messages, regardless of the caller's own session chat id. `with_caller_chat_id`'s implementation is byte-identical before/after R4 in both `PeekQuery` and `ReadQuery` (only doc comments changed). `ensure_owner_only_read_target` (the actual authorization boundary) is untouched by this diff and structurally independent of `participant_filter` — confirmed by req-qa and arch-qa via direct read of `crates/atm-core/src/read/mod.rs:582-599` / `:544` / `async_projection.rs:134`. No cross-agent/cross-team widening anywhere in this diff.

### Important finding — R4's new CLI-peek regression test is vacuous against the scenario it's meant to prove

**QA-4-I1 (rust-qa-agent QA-2)** — `crates/atm-graft-python/tests/test_cli_parity.py:553` (`test_cli_peek_by_listed_id_reaches_bare_agent_message_from_chat`) only sends a **bare-agent-addressed** message (`destination_chat_id=None`) and peeks it by exact ID. Since `matches_participant_filter`'s `To`-direction check never inspects `agent`, only `destination_chat_id` (`crates/atm-core/src/read/filters.rs:56-67`), a bare-agent-addressed message already satisfies `PeekQuery::new`'s restrictive default filter (`Some({agent, chat_id: None})`) with **no fix required at all**. rust-qa-agent proved this by fault-injecting a revert of the `peek.rs` fix: the shipped test still passed. It then constructed a temporary probe sending a **chat-qualified**-addressed message and peeking it by exact ID — that probe correctly failed without the fix and passed with it (reverted after use, final diff confirmed byte-identical to the shipped 4-file/+115/-3 diff). The actual scenario requiring the R4 fix has zero CLI test coverage.

This directly means task requirement (c) — "tests prove (a) for peek the way R2 tests prove it for read" — is **not fully satisfied**: R2's CLI-read test (`test_cli_read_by_listed_id_reaches_bare_agent_message_from_chat`) legitimately proves the fix it guards, because `ReadQuery::new` has no restrictive default to begin with, so a bare-agent-message scenario is the correct evidence there. `PeekQuery::new` DOES have a restrictive default, so peek's parity claim requires the chat-qualified-destination scenario specifically — this was missed.

**Required correction:** add (or replace) a CLI-parity test that sends a message to a chat-qualified destination (`f"{identity}:{chat_id}@{team}"`) and peeks it by exact `--message-id`, confirming it fails without the `peek.rs` fix and passes with it — mirroring what the native unit test in `lib.rs` already does correctly for the native path.

### Important finding — doc-comment defects confirmed by all three reviewers, contradicting the PR body's own disposition claim

**QA-4-I2** — Consolidating quality-mgr's own independent finding (confirmed by all three reviewers): `crates/atm-core/src/read/request.rs:263-268`, `PeekQuery::with_caller_chat_id`'s doc comment is duplicated verbatim (the same 3-line block appears twice in a row). `crates/atm-core/src/read/request.rs:394-395`, `ReadQuery::with_caller_chat_id` has **no doc comment at all**. The implementation bodies of both are confirmed unchanged and correct — this is a doc-only defect.

This directly contradicts the PR body's own QA-3-I1 disposition row (`gh pr view 1299 --json body`): *"Closed: both with_caller_chat_id contracts state that None removes chat scoping and leaves mailbox ownership as the recipient scope."* Only one of the two functions has such text, and even that copy is malformed by duplication.

**Severity adjudication:** req-qa rates this Blocking (a false "Closed" claim on an explicitly-named R4 deliverable in a merge-gating disposition table); arch-qa and rust-qa-agent both rate it Minor (doc-only, no compiled-behavior effect, no `missing_docs` lint enforced in this crate). Quality-mgr adjudicates **Important**: the underlying defect is real and unanimous, and req-qa is right that the PR body's disposition table is factually wrong and must be corrected — but unlike this PR's two prior Blocking calls (QA-1's UDS gap, QA-3's untested behavior-widening), there is zero functional or security risk here, and the fix is a trivial two-line correction plus a PR-body edit. Not merge-blocking on its own, but must be fixed and the PR body corrected before merge.

**Required correction:** remove the duplicate block from `PeekQuery::with_caller_chat_id`, add the same (corrected) doc comment above `ReadQuery::with_caller_chat_id`, and correct the PR body's QA-3-I1 row to accurately describe the state until both comments are actually in place.

### Pre-existing, out-of-scope finding — not a blocker for this PR

**QA-4-N1 (rust-qa-agent QA-1, rated Critical there; quality-mgr downgrades to informational/tracked-follow-up, same disposition pattern as QA-2's ATM-QA-001)** — `crates/atm/src/commands/peek.rs:99` (`resolve_cli_inspection_caller_context` call, `chat_id_override: None` hardcoded) means CLI `atm peek` **bucket scans** (any selection mode other than exact `--message-id`) never apply the caller's session chat scoping at all — `peek.rs` has no `--chat-id` flag and never calls `.with_caller_chat_id(caller_context.caller_chat_id)` in the non-exact-id branch, unlike `read.rs:151-155`. rust-qa-agent proved via fault-injected probe (sent a chat-qualified-addressed message, confirmed `atm peek --all` returns `count: 0` while `atm read --all` on the identical scenario correctly returns `count: 1`) that this silently excludes chat-qualified mail from peek bucket scans. Confirmed present at `origin/develop` (`8845f3b39`) as well — **pre-existing, not introduced by any commit in PR #1299 (R1-R4)**, and out of scope for issues #1297/#1298, which are specifically about idle-pool connections and exact-message-id lookups.

**Disposition: real finding, confirmed, but not a regression this PR introduces and not a blocker for merging the #1297/#1298 fixes.** Recommend: file a tracked follow-up issue to add chat-scoping support (a `--chat-id` override, mirroring `read.rs`) to `atm peek`'s bucket-scan path.

### Checks confirmed directly by quality-mgr

`cargo fmt --all --check` and `cargo clippy --all-targets --all-features -- -D warnings` both clean (run directly, bypassing `just lint`'s wrapper, which fails on both `fmt` and `clippy` targets in this worktree due to a missing `.bootstrap-venv/bin/python` interpreter — a known, pre-existing local-environment gap, not a real code defect). `cargo test --workspace`: 0 failures across every crate. `cargo test -p atm-daemon` explicitly: `version_string_matches_the_binary_contract` passes. rust-qa-agent independently reran the identical full-workspace signal fresh (1841 tests, 0 failed) plus its own fault-injection probes, all reverted cleanly. `gh pr view 1299`: `mergeable: MERGEABLE`, `mergeStateStatus: BLOCKED` only on pending CI; not part of a `gh stack`. PR body correctly states R3 was an intentional behavior change (not a pure refactor), states the peek widening, and states fenix's ruling — the one inaccuracy is the QA-3-I1 disposition row addressed above.

### Merge Readiness
- Status: **not fully ready**. 0 Blocking, but 2 Important findings are real, in-scope gaps against R4's own explicitly-stated deliverables: (1) the CLI-peek regression test doesn't prove the scenario it's meant to guard, (2) the doc-comment fix is incomplete and the PR body overclaims it as done. Both are small, mechanical fixes. Recommend a quick R5 rather than declaring ready.
- QA-4-N1 (peek bucket-scan chat-blindness) is real but pre-existing and out of scope — recommend a separate tracked GitHub issue, not a blocker on this PR.
- All QA-1/QA-2/QA-3 findings remain closed; this round only concerns the new R4 delta.

---
*Posted by quality-mgr. Zero `.triage/`, `.sprints/`, or branch/worktree writes made by quality-mgr.*

randlee:
R5 validation for 467a8e963:

- The replacement other-session CLI exact-ID peek regression passes on the restored R4 hunk.
- Required local fault injection: removing only the R4 peek.rs exact-ID hunk made that test fail with AssertionError: 0 != 1; the hunk was restored before commit. The commit subject records that assertion.
- Rebuilt CLI/daemon plus disposable real-daemon parity suite: PASS, 6 tests.
- cargo fmt --check: PASS.
- cargo clippy --workspace --all-targets -- -D warnings: PASS.
- python3 .just/run_lint.py all: every non-wrapper target and the complete pytests target passed. Its fmt and clippy wrappers failed only because this worktree has no .bootstrap-venv/bin/python; direct cargo replacements above passed.
- git diff c32de7093..467a8e963 --stat: only crates/atm-graft-python/tests/test_cli_parity.py and crates/atm-core/src/read/request.rs, net -2 lines.

The PR body now correctly says the contract appears once on ReadQuery and accurately records the QA4-I1 proof.

randlee:
## Final Quality Report — QA-5 (qa-1299-r5-1788805254)

**PR:** #1299 (`fix/native-loopback-157` → `develop`)
**Head:** `467a8e963` (R5, 2 files +5/-7, on top of QA-4's `c32de7093`)
**Task:** verify QA4-I1 and QA4-I2 are closed exactly as dispatched.
**Verdict:** ✅ **PASS 0/0/0** (0 Blocking, 0 Important, 0 Minor)
**Deliverable completion:** **2/2 (100%)** — both QA-4 Important findings independently confirmed closed by all 3 reviewers (req-qa, arch-qa, rust-qa-agent), plus quality-mgr's own direct verification.

### Reviewer results (3/3 returned, all recheck-only per QA-4's open findings)

| Reviewer | Verdict | Result |
|---|---|---|
| req-qa | PASS (0/0) | QA4-I1 and QA4-I2 both confirmed closed; ownership-scope safety independently re-traced through `read/mod.rs`/`read/filters.rs` — no cross-tenant leak; PR body disposition rows verified indirectly (no `gh` tool in that session) against the underlying code, which supports the claims |
| arch-qa | PASS (0/0) | QA4-I2 confirmed closed; repo-wide grep confirms no duplicate struct/method definitions elsewhere; no RULE-001..013 violation; no speculative/unused code in the 2-file diff |
| rust-qa-agent | PASS (0/0) | Independently rebuilt from a fresh scratch venv, ran the full parity suite (6/6 pass); **independently repeated the fault-injection proof** (reverted the R5 `peek.rs` hunk, reran the renamed test, confirmed `AssertionError: 0 != 1`, restored, reran full suite — all pass); fmt/clippy/full workspace `cargo test` all clean |

### QA4-I1 (Important at QA-4) — confirmed closed by 3 reviewers plus quality-mgr's own independent fault injection

`crates/atm-graft-python/tests/test_cli_parity.py:552-568` was renamed from `test_cli_peek_by_listed_id_reaches_bare_agent_message_from_chat` to `test_cli_peek_by_id_reaches_other_session_message_from_chat`, now sending its fixture message to a **chat-qualified** destination (`f"{identity}:other-session@{team}"`, previously bare-agent) before peeking it by exact `--message-id`.

Quality-mgr independently built the native extension (fresh scratch venv, `maturin develop`) plus real `atm`/`atm-daemon` debug binaries at `467a8e963`, reverted only the R5 fix hunk in `crates/atm/src/commands/peek.rs` (removed the `.map(|query| if self.message_id.is_some() {...} else {query})` closure), reran the renamed test, and confirmed it **fails** with `AssertionError: 0 != 1` — exactly matching the R5 commit message (`fix: cover exact-id peek scope (AssertionError: 0 != 1)`). Restored the fix and reran the full 6-test parity suite: all pass. rust-qa-agent independently repeated this identical fault-injection cycle from its own fresh build, with the same result. This is conclusive: the test now genuinely proves the fix it guards, closing QA-4's vacuous-coverage gap.

### QA4-I2 (Important at QA-4, rated Blocking by req-qa, Minor by arch-qa/rust-qa-agent) — confirmed closed by all 3 reviewers

`crates/atm-core/src/read/request.rs`: `PeekQuery::with_caller_chat_id` (line 267) now has exactly one copy of its doc comment (lines 263-265, duplicate removed); `ReadQuery::with_caller_chat_id` (line 395) now has its own doc comment (lines 391-393), worded distinctly from `PeekQuery`'s but stating the same substance (ownership-only exact-id resolution, `None` clearing the participant filter). Both implementation bodies remain byte-identical to before, confirmed by all reviewers and quality-mgr. `gh pr view 1299 --json body` confirms the QA-3-I1/QA4-I2 disposition row now accurately states "the corrected with_caller_chat_id contract appears once on ReadQuery" and the QA-3-B1/QA-3-I2/QA4-I1 row accurately cites the exact fault-injection detail (`AssertionError: 0 != 1`) — the false-claim basis for req-qa's prior Blocking rating is resolved.

### Findings closed this round

- **QA4-I1** (`crates/atm-graft-python/tests/test_cli_parity.py`) — fixed, verified via independent repeated fault injection by quality-mgr and rust-qa-agent.
- **QA4-I2** (`crates/atm-core/src/read/request.rs`) — fixed, verified via direct read by all 3 reviewers plus quality-mgr.

(Per project convention, quality-mgr does not write `.triage/*.ttl` records — closure is recorded here and in the ATM verdict to fenix for reconciliation.)

### Out of scope, not rechecked this round (per task instruction)

QA-4's pre-existing, out-of-scope finding (rust-qa-agent's Critical, downgraded by quality-mgr at QA-4) — `atm peek` bucket scans never apply session chat scoping — remains untouched by R5's 2-file diff, confirmed by rust-qa-agent. Recommend the previously-suggested tracked follow-up GitHub issue; not part of this PR.

### Checks confirmed directly by quality-mgr

`cargo fmt --all --check` and `cargo clippy --all-targets --all-features -- -D warnings` clean (direct, bypassing `just lint`'s `.bootstrap-venv`-dependent wrapper, a known pre-existing local-environment gap). `cargo test --workspace`: 0 failures. Full `test_cli_parity.py` suite (6/6) via `ATM_CLI_PARITY_CI=1`. `git diff c32de7093..467a8e963`: exactly 2 files, +5/-7, matching the task's stated scope. `gh pr checks 1299`: `Format check`, `Clippy`, `Just lint (ubuntu-latest)`, `Classify CI scope`, and all Hermes ATM wheel/compat jobs green; `Test (macos-latest/windows-latest/ubuntu-latest)` in progress at last check (not a dispatch gate). Not part of a `gh stack`.

### Merge Readiness
- Status: **ready**. 0 Blocking, 0 Important, 0 Minor across all 3 reviewers this round. Both QA-4 Important findings closed with independently-repeated empirical proof, not just code inspection.
- CI green on every completed check; the three platform `Test` jobs were still in progress at last check — required per standing policy before actual merge, not before this verdict.
- Recommend filing the pre-existing `atm peek` bucket-scan chat-scoping gap as a separate tracked GitHub issue, unrelated to this PR's merge.

---
*Posted by quality-mgr. Zero `.triage/`, `.sprints/`, or branch/worktree writes made by quality-mgr.*

Issue #1301, opened 2026-09-07
CLI: global --tcp flag to force TCP loopback to the local daemon (CLI concern, single resolver)


randlee:
## Summary
Rand, 2026-09-07: forcing the same-host transport is a CLI concern, not an application or runtime concern. Add a global `--tcp` flag to the `atm` CLI that makes this invocation talk to the local daemon over TCP loopback instead of the Unix default (UDS). Windows is already TCP-only; there the flag is accepted and a no-op.

## Shape
- One flag on the CLI surface (`crates/atm/src/cli_surface.rs`), alongside the existing caller overrides such as `--as` / `--team`.
- The flag feeds the existing selection in `atm-daemon-client` (`LocalDaemonTransport`, `local_daemon_transport()`), which is the single place a transport is chosen. Flag wins over `ATM_LOCAL_TRANSPORT`; no second parser anywhere.
- No transport selection is added to `atm-http-runtime` or any library crate (PR #1299 R1 did this and it is being removed).
- Documented in the CLI reference and `docs/requirements.md` next to the existing `ATM_LOCAL_TRANSPORT` text.

## Why
An explicit, discoverable flag lets a person or a test run the whole CLI stack over the TCP loopback connector on a Mac without environment plumbing. It is the legitimate form of "test TCP end to end". Rand's rule applies: no code is added unless it is needed and actually used, and every configuration switch doubles the test matrix, so this is one flag, wired to the one existing resolver, with one end-to-end test (a CLI command over `--tcp` against a fixture daemon bound to loopback).

## Not in scope
- No change to the default transport.
- No change to graft / hermes native clients.
- Nothing in PR #1299, which stays the stale-pool and exact-ID fix.

randlee:
HOLD (fenix, Rand 2026-09-07 "if this would bloat cli code, it may not be the best to add"): not dispatched. The need is already met with zero new code by the existing single resolver: `ATM_LOCAL_TRANSPORT=tcp atm <command>` forces TCP loopback for that invocation. Remaining candidate work is documentation only: mention the variable in the CLI reference so it is discoverable. Close this issue unless a concrete need for a flag appears.

randlee:
Closed (Rand 2026-09-07): no code is added that is not necessary and used consistently. The existing ATM_LOCAL_TRANSPORT=tcp path already covers CLI over TCP loopback.

PR #1302, opened 2026-09-07
docs: code addition rule (no code unless necessary and used consistently)

randlee:
Records Rand's 2026-09-07 ruling where every agent reads it: CLAUDE.md gets a Code Addition Rule section; dev-template.xml.j2 and fix-assignment.xml.j2 get a pre-push diff-walk step. Docs-only.

Origin: PR #1299 R1 added transport parsing to atm-http-runtime for test reach; #1301 (CLI --tcp) dropped because ATM_LOCAL_TRANSPORT already covers it.


randlee:
Closed at Rand's direction (misguided). Branch and worktree left in place.

